PR #11, opened 2019-08-28
specify tags for sub-protocols

sticktrk:
It is hard to see if there is a standard for the `tag` parameter per sub-protocol. This is a little more clear and follows the precedent set by the `lnurl-login` section

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -28,6 +28,7 @@ Suppose user has a balance on a certain service which he wishes to turn into an 
 
 User software:
 1. Scans a QR code and decodes a query string.
+	- `tag` with value set to `channelRequest`
 2. Makes an HTTPS GET request to a service.
 3. Gets Json response of form: 
 ```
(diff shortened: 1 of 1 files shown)


btcontract:
There are two types of `lnurl`s:

- type 1 involves two HTTP calls: user scans a QR and obtains a first `lnurl` which precisely does not have a `tag` query parameter, then calls it right away and obtains a Json response which has `tag` as one of fields. `lnurl-channel` and `lnurl-withdraw` both work like this, the reason for splitting calls is to have as small QR as possible (listing incoming channel parameters or too long withdrawal description can make QR unreadable, also this way we are not constrained by QR limits for new sub-protocols).

- type 2 involves a single HTTP call: user scans a QR and obtains an `lnurl` which does have a `tag` query parameter, this is how wallet knows that said `lnurl` must not be called right away but instead some other actions should be performed (such as showing a login window). `lnurl-login` works like this because it only needs to supply `?tag=login&k1=secret`, these can't make QR overly large.

So type 2 is kind of a simplification of type 1 in cases where amount of data is small.

PR #18, opened 2019-11-11
lnurl-pay spec attempt

fiatjaf:


spec.md
```diff
@@ -154,18 +154,45 @@ Note that service will withdraw funds to anyone who can provide a valid ephemera
...
+```
+Additionally, a send dialog must include:
+- Domain name extracted from `lnurl` query string.
+- A way to see full metadata in `text` format.
+
+5. Makes an HTTPS GET request to a service containing `amt`, the amount in MilliSatoshi the user has chosen in the previous screen. May also append a `fromnodes` query parameter with value set to comma separated `nodeId` if payer wishes a service to provide payment routes which start from specified LN `nodeId`s.
+6. Gets a JSON response of form:
+```
+{
+	pr: String, // bech32-serialized lightning invoice with `h` tag set to hash of a whole `metadata` field above
```
baebb: what is the `h` tag?

fiatjaf: `description_hash`

fiatjaf:
Superseded by https://github.com/btcontract/lnurl-rfc/commit/d9f902c9d99c465de24a6165ad9bf17d78d16a74.

PR #19, opened 2019-11-18
Edits for readability

baebb:
make LNURL-pay and withdraw easier to understand for LN services and LN wallets


btcontract:
Thanks!

PR #22, opened 2019-11-20
LNURL-pay add successUrl

baebb:
Adds a `successUrl` to the 2nd LN SERVICE response. eg. mylnservice.com/order/<orderId>

LN SERVICE would create a orderId for the order after LN WALLET has sent how much they would like to pay in `<callback>?amount=<milliSatoshi>&fromnodes=<nodeId1,nodeId2,...>`.

This would be useful for physical or virtual delivery of whatever goods the user paid for.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -245,7 +245,7 @@ Note that service will withdraw funds to anyone who can provide a valid ephemera
 	```
 	{
 		pr: String, // bech32-serialized lightning invoice with h tag set to sha256(utf8ByteArray(metadata))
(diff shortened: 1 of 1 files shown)


fiatjaf:
**OPTIONAL** should be bigger.

# OPTIONAL

baebb:
`optional, but your wallet will suck if you don't support this`

fiatjaf:
Good, now we're offending people on the spec.

It's easy for wallets to implement this, I care more about services returning some stupid unrelated URL when it's not necessary just because they want to fulfill the protocol.

PR #30, opened 2020-01-25
lnurl-withdraw payment attempts can be asynchronous.

fiatjaf:
also remove the multiple payment requests stuff from lnurl-withdraw since no one ever used it and there's MPP now.

### DIFF
--- a/spec.md
+++ b/spec.md
@@ -179,10 +179,10 @@ Today users are asked to provide a withdrawal Lightning invoice to a service, th
 5. Once accepted by the user, `LN WALLET` sends an HTTPS GET to `LN SERVICE` in the form of 
 	
 	```
-	<callback>?k1=<k1>&pr=<lightning invoice>
+	<callback>?k1=<k1>&pr=<lightning invoice, ...>
 	```
 	
-6. `LN SERVICE` sends a `{"status":"OK"}` or `{"status":"ERROR", "reason":"error details..."}` JSON response and then attempts to pay the invoice asynchronously.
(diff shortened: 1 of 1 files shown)


btcontract:
> and then attempts to pay the invoice asynchronously

What does this mean exactly? That OK response from server is independent of actual payment attempt?

> also remove the multiple payment requests stuff from lnurl-withdraw since no one ever used it and there's MPP now

MPP does not replace this entirely. Imagine receiver has 2 private channels where one can receive 100 sat and the other one 10 sat. Receiver can then create 2 invoices where the first one expects 100 sat and hints first channel only and second one expects 10 sat and hints second channel only. With such approach there is no need for guessing on sender side and delivery chances are better.

fiatjaf:
> What does this mean exactly? That OK response from server is independent of actual payment attempt?

Yes. Some people have been confused by this.

> MPP does not replace this entirely. Imagine receiver has 2 private channels where one can receive 100 sat and the other one 10 sat. Receiver can then create 2 invoices where the first one expects 100 sat and hints first channel only and second one expects 10 sat and hints second channel only. With such approach there is no need for guessing on sender side and delivery chances are better.

I imagine this would be a ton of work to do and no service will ever care to implement just because of this corner case. Also wallets probably won't bother also,  it would be too complicated for them. Perhaps not for BLW as it has all channel data at arm's reach.

btcontract:
> Yes. Some people have been confused by this.
OK

> I imagine this would be a ton of work to do and no service will ever care to implement just because of this corner case. Also wallets probably won't bother also, it would be too complicated for them. Perhaps not for BLW as it has all channel data at arm's reach.

I intend to implement this both on service and BLW side, please leave this option as is and we can merge.

PR #37, opened 2020-02-29
Allow embedding lnurl into tags, define tagged field identifier

akumaigorodski:


takinbo (approved):
I had a little difficulty accepting the usefulness of this change but then I had an idea that might provide justification for this:

Imagine a merchant has provided an invoice for the purchase of a certain item: let's assume it's an invoice to top up a prepaid electricity meter with $100 worth of energy. Although the invoice is for a one-time payment, embedding a lnurl in the invoice can give the wallet the ability to save a reference to that product (lnurl-pay market) so it can be repurchased in the future.

ACK!

fiatjaf:
Can we merge this?

PR #39, opened 2020-04-01
Update lnurl-withdraw.md

jackeveritt:
Should 'q' be added here


fiatjaf:
No. What is this `q`? Any service is free to add any querystring parameter they want, but it's not in the spec.

PR #43, opened 2020-05-07
Specify k1 requirements on server side

btcontract:


fiatjaf:
I have nothing against this, but I want to understand what are the risks associated with k1 reuse or the server accepting unexpected k1s.

btcontract:
Without these restrictions an already used k1 and signature can be reused without having a private key. While very unlikely that an attacker can get these it's still a security degradation.

PR #47, opened 2020-07-05
Add auth test vector

akumaigorodski:


fiatjaf:
This makes no sense to me. Where is the linkingKey?
Also we need the seed from which the hashingKey is derived in the test vector!

theborakompanioni:
I also think it'd be great if those values can be found in the spec.

Three small suggestions come to mind:
- like @fiatjaf said i think it'd be good if the wallet seed is and plain `linkingKey` given
- what about using `example.com` (or a domain the project is in control of) for example urls?
- also, a human readable path e.g. `full path: m/138'/1603989739'/682320451'/2081389135'/1986563837'` (for given `Vector(3751473387, 2829804099, 4228872783, 4134047485)`)

hsjoberg:

PR #48, opened 2020-07-05
Add random nonce requirement for a callback

akumaigorodski:


fiatjaf:
I think this should remain an open PR forever and never be merged.

PR #49, opened 2020-07-16
LNURL Allowance

andrerfneves:
## Notes

After much deliberation we're putting forward the first version of what we believe to be a good, yet very minimal, spec for a LNURL Allowance flow. On our side, the idea behind LNURL Allowances started with [Christian's Lightning Network Gaming Hackathon demo](https://www.youtube.com/watch?v=ZNKwL0zloN8), but it was also explored on [Lightning Joule's Allowances feature](https://medium.com/@wbobeirne/introducing-joule-allowances-2b08bec75e3a). Since then we've been trying to create a more standardized approach to this new feature that would fit into the larger LNURL group of Lightning UX Specifications for Wallets and Services.

This PR is the first iteration of it after long discussions with @fiatjaf and @mandelmonkey. We're now looking for feedback/contributions on it, and hope to have some demo implementations on Wallets/Services in the coming weeks. On [ZEBEDEE](https://zebedee.io)'s side we're really excited about what this brings to the usability of Lightning in gaming environments. We plan to add this feature to both our Android/iOS mobile wallet, as well as our set of APIs, thus allowing Game Developers building on ZEBEDEE's set of APIs and SDKs to make use of this feature as well.

- e.g. being able to scan a single QR at the beginning of your `interaction/gameplay`, stipulate a `spending amount` and then enjoy the experience without the burden of scanning QR codes for every interaction.

That being said, LNURL Allowance is meant to be a service-type-agnostic solution and does not ONLY solve our problems. We have seen interest to introduce similar functionality to Lightning services of various other types/characteristics.

## Additional

I've added a `Known Issues` section at the very end given that maintaining WebSocket connections active on Mobile devices can be rather tricky.

I will also be adding a diagram [similar to this one](https://i.imgur.com/ZV0IT1g.png) to the PR once it's been more finalized/worked out/modified due to community feedback. The diagram linked above was the very first iteration of this specification, and is not meant to be followed. These diagrams have helped me explain/illustrate the concept more clearly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,12 +4,13 @@
 
 `LNURL` is a bech32-encoded HTTPS/Onion query string which is supposed to help payer interact with payee and thus simplify a number of standard scenarios such as:
 
-- Requesting incoming channels ([LNURL-channel](lnurl-channel.md)) 
+- Requesting incoming channels ([LNURL-channel](lnurl-channel.md))
 - Logging in ([LNURL-auth](lnurl-auth.md))
 - Withdrawing funds ([LNURL-withdraw](lnurl-withdraw.md))
 - Paying for a service ([LNURL-pay](lnurl-pay.md))
+- Pairing Wallet <-> Service through allowances ([LNURL-allowance](lnurl-allowance.md))
 
-An example `LNURL`: 
+An example `LNURL`:
 > https://service.com/api?q=3fc3645b439ce8e7f2553a69e5267081d96dcd340693afabe04be7b0ccd178df
 
 would be bech32-encoded as:
@@ -37,11 +38,11 @@ In Scala:
 ```scala
 import fr.acinq.bitcoin.Bech32
 val bech32lnurl: String = "LNURL1DP68GURN8GHJ7UM9WFMXJCM99E3K7MF0V9CXJ0M385EKVCENXC6R2C35XVUKXEFCV5MKVV34X5EKZD3EV56NYD3HXQURZEPEXEJXXEPNXSCRVWFNV9NXZCN9XQ6XYEFHVGCXXCMYXYMNSERXFQ5FNS"
-  
+
 val (hrp, dataPart) = Bech32.decode(bech32lnurl)
-  
+
 val requestByteArray = Bech32.five2eight(dataPart)
-  
+
 new String(requestByteArray, "UTF-8") // https://service.com/api?q=3fc3645b439ce8e7f2553a69e5267081d96dcd340693afabe04be7b0ccd178df
 ```
 
@@ -55,3 +56,4 @@ If you have any questions about implementing LNURL as a wallet or service, join 
 2. [LNURL-auth](lnurl-auth.md)
 3. [LNURL-withdraw](lnurl-withdraw.md)
 4. [LNURL-pay](lnurl-pay.md)
+4. [LNURL-allowance](lnurl-allowance.md)
--- a/README.md
+++ b/README.md
@@ -56,4 +56,4 @@ If you have any questions about implementing LNURL as a wallet or service, join 
 2. [LNURL-auth](lnurl-auth.md)
 3. [LNURL-withdraw](lnurl-withdraw.md)
 4. [LNURL-pay](lnurl-pay.md)
-4. [LNURL-allowance](lnurl-allowance.md)
+5. [LNURL-allowance](lnurl-allowance.md)
--- a/lnurl-allowance.md
+++ b/lnurl-allowance.md
@@ -1,6 +1,6 @@
 # LNURL-allowance
 
-The LNURL Allowance scheme allows a SERVICE and a WALLET to establish a `pairing` over a WebSocket connection. The idea is that SERVICE will provide some recommended criteria (pairing amount and pairing time) to WALLET and once displayed and accepted by the user of WALLET, will allow for WALLET to pay/receive funds to/from SERVICE without user interaction or QR Code scanning.
+The LNURL Allowance scheme allows a `LN SERVICE` and a `LN WALLET` to establish a `pairing` over a WebSocket connection. The idea is that `LN SERVICE` will provide an LNURL that returns some parameters (including a socket URL) to `LN WALLET`. `LN WALLET` will diplay a pairing dialog and once accepted by the user, `LN WALLET` will start a socket connection to `LN SERVICE`. This will allow for `LN WALLET` to pay/receive funds to/from `LN SERVICE` without user interaction or QR Code scanning, up to the established allowance amount.
 
 ## Scan QR code to establish allowance/pairing
 
@@ -16,10 +16,9 @@ The LNURL Allowance scheme allows a SERVICE and a WALLET to establish a `pairing
 {
     socket: String, // the URL from LN SERVICE which will accept the WebSocket connections from the WALLET
     k1: String, // random or non-random string to identify the WALLET when receiving connection to socket
-    requestAllowanceTime: Seconds, // amount of time the SERVICE recommends WALLET to ask user for this allowance/pairing
-    requestAllowanceAmount: Seconds, // amount of seconds the SERVICE recommends WALLET to ask user for this allowance/pairing
-    image: Base64, // base64-encoded SERVICE image for WALLET to display to user when establishing connection
-    description: String, // SERVICE description that WALLET should showcase to user
+    recommendedAllowanceAmount: Millsatoshis, // amount of satoshis the SERVICE recommends WALLET to ask user for this allowance/pairing
+    image: Base64, // base64-encoded LN SERVICE image for LN WALLET to display to user when establishing connection
+    description: String, // LN SERVICE description that LN WALLET should showcase to user
     tag: "allowanceRequest" // type of LNURL
 }
 ```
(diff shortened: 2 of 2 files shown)


lnurl-allowance.md
```diff
@@ -0,0 +1,197 @@
...
+
+2. `LN WALLET` makes a GET request to `LN SERVICE` using the decoded LNURL.
+
+3. `LN WALLET` gets JSON response from `LN SERVICE` of form:
+
+```
+{
+    socket: String, // the URL from LN SERVICE which will accept the WebSocket connections from the WALLET
+    k1: String, // random or non-random string to identify the WALLET when receiving connection to socket
+    recommendedAllowanceAmount: Millsatoshis, // amount of satoshis the SERVICE recommends WALLET to ask user for this allowance/pairing
```
baebb, line 19: could this be changed to `minAllowanceAmount` and `maxAllowanceAmount`, like with LNURL-withdraw?

README.md
```diff
@@ -55,3 +56,4 @@ If you have any questions about implementing LNURL as a wallet or service, join
 2. [LNURL-auth](lnurl-auth.md)
 3. [LNURL-withdraw](lnurl-withdraw.md)
 4. [LNURL-pay](lnurl-pay.md)
+4. [LNURL-allowance](lnurl-allowance.md)
```
baebb, line 59: this should be 5.

andrerfneves: Good catch! Thanks

lnurl-allowance.md
andrerfneves: With the setup we put forward, the WALLET is constantly telling the SERVICE their `balance` through the `balance` message (which also serves as a PING). The reason we don't set a `maximum` amount is that if the WALLET runs out of funds, we'd need to have another message type to `requestNewAllowance` sort of thing. If the WALLET/user would like to `increase` their balance, we can let the WALLET handle that, and provide easy-to-use buttons/UI to handle the `add 5000 sats to balance` sort of thing. The SERVICE just needs to receive that value and know whether the balance is enough for given `interaction/gameplay/whatever`. By setting an upper bound limit, we also increase the amount of interaction the user would have to go through to `increase it`.

The idea behind a `recommendedAllowanceAmount` is that the SERVICE can let the WALLET know what they recommend: e.g. in a game scenario, let's say you usually/likely need 1000 sats to complete the given mission, therefore the SERVICE can suggest/recommend 1000 sats, and you as the WALLET owner can complete given mission before having to `increase balance`. If we don't set a `minimum/recommended` value, the WALLET could default to 50 sats for example, and then the user has to stop and increase their balance right away. The goal for this spec is to minimize the burden around user interactions for little microtransactions like this.

I'm not against the idea of recommending a lower bound and an upper bound, but it should be a recommendation, not a hard limit `maxAllowanceAmount` if that makes sense?

baebb: Thanks for explaining

fiatjaf: Now I'm thinking that since it is just a suggestion to the wallet maybe it could have a `minAllowanceAmount`/`maxAllowanceAmount` just to be similar to the other protocols, but this is a minor thing.

lnurl-allowance.md
```diff
@@ -0,0 +1,197 @@
...
+    image: Base64, // base64-encoded LN SERVICE image for LN WALLET to display to user when establishing connection
```
fiatjaf, line 20: What about the image mimetype? I think sometimes it is necessary to know that beforehand.

lnurl-allowance.md
```diff
@@ -0,0 +1,197 @@
...
+    description: String, // LN SERVICE description that LN WALLET should showcase to user
```
pseudozach, line 21: should this field be more like lnurl-auth rather than lnurl-withdraw? 
Maybe what's shown to the user needs to be pulled from domain to ensure the request comes from the actual website. 
otherwise anyone can put a description with a popular game and immediately withdraw all the allowance.

jarret:
Ross asked me to comment here, so I am doing so. I don't consider myself a stakeholder in LNURL and well-versed in its project goals, though this is probably a concept NACK.

I previous gave my feedback to @andrerfneves and @fiatjaf  in-depth in a discussion in t.me/moneysocket

In summary, I think there are shortcoming in:

1) not accommodating for the need for SERVICEs to connect to WALLETs in addition to the setup of WALLETs connecting to SERVICEs outlined here. Many WALLETs are full nodes with fixed infrastructure and likewise, potential SERVICEs working off of mobile, dedicated hardware or browser runtime instances are quite imaginable.

2) covering the needs for overcoming challenges of network firewalls on both ends - not all SERVICE and WALLET use cases are able to provide a full, permanent, DNSed, routable websocket server and by structuring it one way, you are leaving out those use cases. Considering compatibility with other interconnects than websocket (WebRTC, bluetooth, NFC, etc.) as well as websocket relay infrastructure might be advisable.

3) the proposal is built around giving permission to a SERVICE you don't control to move money. I suggest a better paradigm would enable you to give permission from YOUR spending wallet to a different piece of software YOU control to automate purchasing on your behalf. This setup would make the web service you are using a WALLET (that your software automation is deciding to send money to) rather than a SERVICE (that is pulling money from your wallet at its prerogative).

Also, I have previously observed in discussion that the REST call is unnecessary since that same data could be exchanged via websocket. This would avoid needing to provide REST client/server infrastructure to implement this proposal.

lnurl-allowance.md
andrerfneves: I'm not against having an upper and a lower bound defined by the SERVICE. Some questions I have:

1 - Can the WALLET put an allowance value `larger than` the `maxAllowanceAmount`?
2 - Can the WALLET put an allowance value `smaller than` the `minAllowanceAmount`?

I understand these are recommendations, but should we have them be more than that? Such that the WALLET has to perform the allowance at the `very least` on the `minAllowanceAmount` and at the `very most` the `maxAllowanceAmount`? Given the WALLET is the one that handles the actual management of the funds in the `allowance`, this shouldn't be an issue either way.

lnurl-allowance.md
andrerfneves: Yeah good point. Should we enforce a mimetype, or do we want to add another property for it? At that point, we may want to follow the approach for the multi-property `content metadata` property present in LNURL Pay?

lnurl-allowance.md
andrerfneves: Yes I agree with this. As you can see, the WALLET is responsible for doing exactly what you mentioned: [you can see that here](https://github.com/btcontract/lnurl-rfc/pull/49/files/528191fe8c39036e12c6e9605791178704698d7c#diff-0a4dd2f0bb49676afea53bb019139103R40)

andrerfneves:
> 1 - not accommodating for the need for SERVICEs to connect to WALLETs in addition to the setup of WALLETs connecting to SERVICEs outlined here. Many WALLETs are full nodes with fixed infrastructure and likewise, potential SERVICEs working off of mobile, dedicated hardware or browser runtime instances are quite imaginable.

LNURL Allowance is not solving for SERVICEs that have to connect to WALLETs. Actually none of the LNURL protocols allow for that. They are simply a set of messaging standards such that both sides (SERVICE and WALLET) can understand one another's payloads. Every LNURL interaction begins at the WALLET side. The SERVICE provides a QR code, but the WALLET is who initiates the interactions. 

> 2 - covering the needs for overcoming challenges of network firewalls on both ends - not all SERVICE and WALLET use cases are able to provide a full, permanent, DNSed, routable websocket server and by structuring it one way, you are leaving out those use cases. Considering compatibility with other interconnects than websocket (WebRTC, bluetooth, NFC, etc.) as well as websocket relay infrastructure might be advisable.

We've discussed this before, but I'm in the mindset that using multiple messaging protocols is NOT in our best interest. Even having support for them could be tricky. That means every SERVICE or WALLET that wishes to implement LNURL Allowance will have to spend time testing/fixing/maintaining and discovering/patching edge cases for each of these protocols. Additionally, NFC and Bluetooth are `physical distance` and `hardware` dependent. WebSockets are extremely well established in the web industry and have support in most modern web clients/server frameworks and environments. No need to over complicate here.

I'd also add that most SERVICEs are cloud-hosted / self-hosted standalone boxes/servers/hardware. While WALLETs are likely `on-the-go` mobile devices (as well as web clients of course). Without the overhead of an always-available intermediary software like you mention below, I fail to see how WALLETs would be `firewalled open to accepting connections from any SERVICE that wishes it`. 

> 3 - the proposal is built around giving permission to a SERVICE you don't control to move money. I suggest a better paradigm would enable you to give permission from YOUR spending wallet to a different piece of software YOU control to automate purchasing on your behalf. This setup would make the web service you are using a WALLET (that your software automation is deciding to send money to) rather than a SERVICE (that is pulling money from your wallet at its prerogative).

Your approach has a large requirement which is this `different piece of software you control`. What exactly is that? Is that cloud hosted or do you need to host it yourself? Are there specific hardware requirements? Does the WALLET provider build and maintain that for its users? Also, help me understand: if you/your wallet are simply giving spending permission to this `secondary software that you control` to do the spending on your behalf, how is that different? Isn't this just another `step` that's added? Do you `pay` this secondary software the total allowance amount, and then the `balance` and `payments` are handled by that software? What sort of security measures would this `secondary software` have that would be an added bonus, and which simply cannot be part of the WALLET software itself? Apart from the example you've mentioned a few times of the `pay-per-livestream` scenario, I have yet to find a concrete use-case where a SERVICE would initiative the interaction with the WALLET. At that point, shouldn't there be some form of business logic that allows a external SERVICE to possibly send a push notification / alert / webhook to request action on the WALLET side (essentially `asking for the WALLET to start the interaction`)?

> 4 - Also, I have previously observed in discussion that the REST call is unnecessary since that same data could be exchanged via websocket. This would avoid needing to provide REST client/server infrastructure to implement this proposal.

While this is technically true, it also goes against the current paradigm for all LNURL protocols. The first GET call is for a `give me the details` and then follows a set of actions. We want LNURL Allowance to be low in development efforts and keep parity with the other LNURL protocols. It also allows for a clean way to provide `properties`/`details` of this Allowance, same way it does for Pay and Withdraw.

lnurl-allowance.md
fiatjaf: I don't know. That property is only there in that specific format because it needed to be hashed and included in the invoice description. There are nicer ways to do it without that constraint.

Anyway, I'm implementing this assuming images will be PNG, but I think this only matters for Telegram, and we can add an optional parameter later.

lnurl-allowance.md
fiatjaf: Yes, I think the wallet SHOULD enforce that on the user, but it remains the decision of the wallet + user, in whatever decision-making process these two agree on.

rafaelpac:
A couple months ago I was thinking about something similar to this.
I have it half-written here somewhere, I called it **LNURL-pre-pay**.

I was thinking about IoT devices that need to charge a price, but the price is not yet known beforehand.
Examples: gasoline pump, rental bicycle... (any bulk purchase or rental machine service).
Instead of maintaining a connection between wallet and service, the wallet would know that this is a pre-payment and it would:
- pay a regular LNURL-pay payment, plus
- send, as the comment string, an amountless invoice (or its own LNURL-pay static string, Zebedee user id, lntxbot user...).

The service now receives that maximum amount and knows that it can use it.
There could be a suggested amount provided by the service, or max/min, as discussed above.
At the end, any change left is paid back to the wallet.
The end happens either when the funds end or when the user activates the end (the user signals to the service that it is enough of that product, service...).

Reading this proposal, I now just had the idea on embedding in it an option to quickly add more funds to the pre-payment, to prevent the service from stopping abruptly in case it uses all the funds.

hsjoberg:

danarchos:
A new PR never seemed to open for this, what is the current status on allowances? @andrerfneves

andrerfneves:
We've yet to re-draft this, but someone else needs to champion it atm.

dni:
:eyes:

PR #50, opened 2020-08-03
Support for LNURL Pay Comment

andrerfneves:
## Notes

This allows for an LN WALLET user to pass a custom `comment` property to an LN SERVICE that accepts it. This can have multiple use-cases --> better donation flow is an easy one to visualize. The entity donating gets to send a message, whether their name/contact/random text and the issuer gets to be paid to read/voice/receive that message, for example streamers.

This is completely additional/add-hoc, so current implementations are not affected by this. Any services and wallets that wish to support it do need to add some more business logic.

### DIFF
--- a/lnurl-pay.md
+++ b/lnurl-pay.md
@@ -14,7 +14,7 @@
         maxSendable: MilliSatoshi, // max amount LN SERVICE is willing to receive
         minSendable: MilliSatoshi, // min amount LN SERVICE is willing to receive, can not be less than 1 or more than `maxSendable`
         metadata: String, // metadata json which must be presented as raw string here, this is required to pass signature verification at a later step
-        commentAllowed: Number, // number of characters accepted for the `comment` query parameter on subsequent callback, defaults to 0. (no comment allowed)
+        commentAllowed: Number, // optional number of characters accepted for the `comment` query parameter on subsequent callback, defaults to 0 if not provided. (no comment allowed)
         tag: "payRequest" // type of LNURL
     }
     ```
--- a/lnurl-pay.md
+++ b/lnurl-pay.md
@@ -24,7 +24,7 @@
     {"status":"ERROR", "reason":"error details..."}
     ```
 
-    `metadata` - json array must contain one `text/plain` entry, all other types of entries are optional
+    `metadata` json array must contain one `text/plain` entry, all other types of entries are optional:
 
     ```
     [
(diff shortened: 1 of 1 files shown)


lnurl-pay.md
```diff
@@ -7,24 +7,25 @@
...
 3. `LN WALLET` gets Json response from `LN SERVICE` of form:
-    
+
     ```
     {
         callback: String, // the URL from LN SERVICE which will accept the pay request parameters
         maxSendable: MilliSatoshi, // max amount LN SERVICE is willing to receive
         minSendable: MilliSatoshi, // min amount LN SERVICE is willing to receive, can not be less than 1 or more than `maxSendable`
         metadata: String, // metadata json which must be presented as raw string here, this is required to pass signature verification at a later step
+        commentAllowed: Number, // number of characters accepted for the `comment` query parameter on subsequent callback, defaults to 0. (no comment allowed)
```
btcontract, line 17:
```suggestion
        commentAllowed: Number, // optional number of characters accepted for the `comment` query parameter on subsequent callback, defaults to 0 if not provided. (no comment allowed)
```

andrerfneves: Looks good

hsjoberg:
I like this. It makes lnurl-pay more interactive, which is something we should strive for now when we finally have the opportunity.

One idea would be to make this more dynamic.

Two things:
* It is currently limited to one input, there are lots of scenarios where it could be useful to have multiple inputs (shipping, account registration (username, password) etc).
* Perhaps it should be possible to name what the field is about, this let's the wallet have nicer UX instead of simply just "comment". Like yalls could name the field "Message" and a shop could name the field "Additional notes".

You could make it an array of

```Typescript
interface MetaData {
  name: string;
  numChars: number; 
}
```

Wallet receiving an empty array (or key doesn't exist) would indicate no comment/metadata.

Just my thoughts, I love this idea and it's something I've been thinking about as well.

fiatjaf:
What if there is just an arbitrary text field, and if in the future apps and wallets want more data to be sent wallets can display a form and send data as JSON or something else and services will be expecting that already (if that future ever happens).

That would make the protocol simple and practical now, with just an added text field, but in the future increased flexibility (and complexity) could just be added with this "soft-fork".

fiatjaf:
Another thing that could be done here is to send this data with a POST. Without breaking backwards-compatibility, services supporting the comment thing could accept either GET or POST requests, while a wallet that supporting the comment thing could send data as POST if there was a comment and GET if not. I think this works and nothing breaks.

PR #51, opened 2020-08-09
simplest redirect-based lnurl-menu proposal.

fiatjaf:
This is the easiest to implement LNURL subprotocol ever, and probably the most flexible and powerful.

Leaving this here as an easy way to reference this idea already exposed and discussed on https://t.me/lnurl.

(Besides being very simple, it still has a lot of room for other complex things to be added to it.)

### DIFF
--- a/lnurl-menu.md
+++ b/lnurl-menu.md
@@ -1,0 +1,33 @@
+# LNURL-menu
+
+## Navigating through a service
+
+Provides a way for services to supply wallets of other LNURL options they have to offer.
+
+### Wallet to service interaction flow:
+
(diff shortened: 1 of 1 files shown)


arcbtc:
Could LNURL-auth be used before showing the user the pay/withdraw/channel options?

arcbtc:
Or a paywall, then you could make wallet native gambling. I pay, then get a withdraw

takinbo:
@arcbtc that is already possible; unless what you mean is that within the lnurl-withdraw/pay/channel workflow, you can additionally authenticate the user. I can see how that would be useful but that should probably be in a new issue.

hsjoberg:

PR #52, opened 2020-08-31
Add auth v2

akumaigorodski:


fiatjaf:
Much better than adding "auth v2" is to find everybody who is using "auth v1" and tell them to change. Then delete "auth v1" from the repository. Can we do that instead?

Otherwise this will be very confusing.

btcontract:
Even at current usage levels it's next to impossible to orchestrate a transition flawlessly, the best we can do is convince wallets and services to support both methods.

fiatjaf:
That is also impossible to do flawlessly and much more likely to cause confusion and frustration than anything else.

btcontract:
I don't see how that is the case sorry.

BitcoinErrorLog:
Do both? Afaik the LNURL community is pretty tight. Contacting the people that have implemented shouldn't be hard, you probably already have a list and it would take one day.

btcontract:
Yes, doing both is exactly a plan. V2 is designed such that it can be added on top of V1 on both client and server side without breaking anything.

fiatjaf:
@BitcoinErrorLog is absolutely correct.
And doing both is terrible and will kill lnurl-auth.

BitcoinErrorLog:
Don't let me have too much weight, my only exposure to LNURL since I left Bitrefill is basically LNmarkets. But maybe they are the main user that matters?

fiatjaf:
What is the main reason for a v2 after all?
Some subtle extra features surely can't justify such a change.

PR #56, opened 2020-10-17
Add domain name guides

akumaigorodski:
Clarifies https://github.com/btcontract/lnurl-rfc/issues/55

### DIFF
--- a/lnurl-auth.md
+++ b/lnurl-auth.md
@@ -14,6 +14,8 @@ Once `LN SERVICE` receives a call at the specified `LNURL-auth` handler, it shou
 
  `LN SERVICE` should carefully choose which subdomain (if any) will be used as LNURL-auth endpoint and stick to chosen subdomain in future. For example, if `auth.site.com` was initially chosen then changing it to, say, `login.site.com` will result in different account for each user because full domain name is used by wallets as material for key derivation.
 
+ `LN SERVICE` should consider giving meaningful names to chosen subdomains since `LN WALLET` may show a full domain name to users on login attempt. For example, `auth.site.com` is less confusing than `ksf03.site.com`.
+
 ### Key derivation for Bitcoin wallets:
 
 Once "login" QR code is scanned `linkingKey` derivation in user's `LN WALLET` should happen as follows:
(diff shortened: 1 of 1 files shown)


dpad85:
Why not let the `service` define a `domain` attribute and add it to the `LNURL` parameters along with `k1`? Something like":

- if `domain` is present and valid(*), the wallet MUST use `domain` to sign `k1`;
- if `domain` is present but not valid, the wallet MUST reject the authentication attempt;
- if `domain` is not present, the wallet behaviour is not defined (so like it is today).

(*) validity rules could be inspired by [cookies rfc 6265 §4.1.2.3](https://tools.ietf.org/html/rfc6265#section-4.1.2.3) and [rfc 6265 §5.2.3](https://tools.ietf.org/html/rfc6265#section-5.2.3).

This would make implementation simpler for both the wallet and the service, and ensure that LNURL accounts do not depend on the wallet implementation or the service API host.

fiatjaf:
I like this, but then a subdomain can claim authority over a domain up in the hierarchy? What if mysite.com.br claim authority over com.br?

dpad85:
> But then a subdomain can claim authority over a domain up in the hierarchy? What if mysite.com.br claim authority over com.br?

The service would not claim authority by using an improper domain. It would not give the service access to anything. This would just make a wallet generate a signing key using some public suffix. This public key would not be used by any other services, because they are properly configured.

In your example, `sub.site.com` is properly configured and would ask the wallet to generate a signing key with domain `sub.site.com`. Which the wallet may or may not do! The service has no way to know. If someone signs in with a key for `site.com` or `foo.bar.com`, it would just work. The only issue would be that the public key would also be used on other services, and one could reconcile data from those services and know that's the same user. It's about privacy.

In any case, if a service is misconfigured and uses a public suffix domain like `com.br` or a domain unrelated to LNURL host, the wallet should reject the attempt. That's why I was mentioning the cookies rfc for validation rules (same issue but worse since a misconfigured cookie can leak valuable information). Quoting section 4.1.2.3:

>    NOTE: For security reasons, many user agents are configured to reject
   Domain attributes that correspond to "public suffixes".  For example,
   some user agents will reject Domain attributes of "com" or "co.uk".
   (See Section 5.3 for more information.)

fiatjaf:
I don't know, the wallet can already decide today what they will use to generate the keys. The protocol cannot enforce any of these choices. The wallet can even decide to use the same key on all websites and that's fine -- except for loss of privacy and compatibility.

Introducing a parameter that may be ignored or rejected by each wallet according to arbitrary criteria does nothing to improve that.

PR #59, opened 2021-02-22
Proposal: LNURL-deposit

thorie7912:
This is a rough sketch of an idea to provide a URL-based registration process to allow wallets and services to support asynchronous direct deposit payments.


thorie7912:
Instead of being it's own whole spec, seems like this could be a part of LNURL-withdraw as a `recurring: true` option so you can reuse the callback inside the JSON.

thorie7912:
Discussed in Telegram, I think a LNURL-withdraw extension to support static or reusable callback endpoints makes more sense for now. 
We can have a layer for managing other "push payments" maybe in the future but for now, it's less work for everyone to just implement an extension - even for server-based wallets like BTCPay.

fiatjaf:
For reference: https://t.me/lnurl/9233

Issue #14, opened 2019-10-12
Automatic withdrawals


Kixunil:
Summary
-------------

I propose to specify a new protocol for automated user-friendly withdrawals. It's a smarter way to implement invoice-less sending.

Motivation
-------------

It often occurs that some kind of automated system of the payer can determine the obligation to pay. Examples include: ATMs (based on user inserting money), exchanges (based on automatic buys facilitating DCA), payroll software...

Currently the process is as follows:

* The service must somehow notify the user about the balance (this is often, but not always during user interaction)
* The user is requested to enter an invoice
* The user manually creates the invoice and enters it into the system somehow
* The system attempts to pay the invoice
* If the attempt fails, refund needs to be processed
* In theory, channel open with push (Turbo or not) could be offered via LNURL, which would require further manual interaction. I'm not aware of any service that provides this currently.

This process has several serious drawbacks:

* It requires many manual steps, which isn't particularly user-friendly
* It may be confusing for newbies
* Amount must be correctly manually entered, amount-less invoices are insecure
* The server has no way of determining if the wallet is compatible with LNURL protocol

As a solution I propose a new sub protocol to LNURL.

All this is based on real-life experience with various LN services, first-hand experience with customers and baristas in Paralelna Polis and discussion with Karel Kyovsky, CEO of General Bytes - Bitcoin ATM manufacturer.

Design goals
-----------------

* Sending over existing LN channels is preferred
* The user should not be bothered with deciding which method to use
* Provide a good fallback for scenarios where LN payment fails
* The protocol should be well-usable for both mobile and "server" (always-online routing node) wallets
* Specify minimum (but extensible) protocol according to current needs. The intent is to not over-specify

Disclaimer
-------------

I've read the documentation of Firebase and don't have a personal experience with it. It may happen that this specification is lacking something to actually allow push notifications.

Specification
----------------

The response message from the server is:

```jsonc
{
    "tag": "autoWithdraw",
    "k1" : "<RANDOM SECRET>",
    "capabilities": {
        "websocket": "wss://example.com/ws", // Websocket notification endpoint
        "fcm": "https://example.com/fcm", // Push notification registration endpoint
        "openChannel": { allowed: true, "uri": "<LN NODE URI>", "min" 100000, "fee": 1000, "spendUnconfirmedPush": true } // the server supports opening a channel. The fee is flat in satoshis
    },
    "callback": "https://example.com/withdraw",
[...]

PR #60, opened 2021-03-13
introduce `balanceCheck` and `balanceNotify`.

fiatjaf:
As discussed here: https://t.me/lnurl/9729


lnurl-withdraw.md
```diff
@@ -13,12 +13,13 @@ Today users are asked to provide a withdrawal Lightning invoice to a service, th
...
     {
+        tag: "withdrawRequest", // type of LNURL
         callback: String, // the URL which LN SERVICE would accept a withdrawal Lightning invoice as query parameter
         k1: String, // random or non-random string to identify the user's LN WALLET when using the callback URL
-        maxWithdrawable: MilliSatoshi, // max withdrawable amount for a given user on LN SERVICE
         defaultDescription: String, // A default withdrawal invoice description
-        minWithdrawable: MilliSatoshi // An optional field, defaults to 1 MilliSatoshi if not present, can not be less than 1 or more than `maxWithdrawable`
-        tag: "withdrawRequest" // type of LNURL
+        maxWithdrawable: MilliSatoshi, // max withdrawable amount for a given user on LN SERVICE
+        minWithdrawable: MilliSatoshi, // An optional field, defaults to 1 MilliSatoshi if not present, can not be less than 1 or more than `maxWithdrawable`
```
Kixunil, line 21: Should be allowed to be zero. The use case is in ATMs which want to make sure users wallet supports this before allowing to insert money. This avoids having to process returns of money.

lnurl-withdraw.md
```diff
@@ -36,13 +37,39 @@ Today users are asked to provide a withdrawal Lightning invoice to a service, th
...
 6. `LN SERVICE` sends a `{"status": "OK"}` or `{"status": "ERROR", "reason": "error details..."}` JSON response and then attempts to pay the invoices asynchronously.
 7. `LN WALLET` awaits for incoming payment if response was successful.
 
 Note that service will withdraw funds to anyone who can provide a valid ephemeral `k1`. In order to harden this a service may require autorization (LNURL-auth, email link etc.) before displaying a withdraw QR.
 
+### `balanceCheck`
+
+This _optional_ property is meant to enable automatic delivery of funds from services to wallets. It works as follows:
+
+  * If `LN WALLET` supports the `balanceCheck` scheme and sees a `balanceCheck` property from a withdraw it performs, it may store that URL locally and call it later (for example, at its next startup, or after 24 hours).
```
Kixunil, line 55: To complement the above, I'd add: "If the withdrawable amount is zero and the wallet does **not** support `balanceNotify`, then the wallet SHOULD poll the service once per minute for the following 30 minutes."

The idea is that zero means it's expected to go above zero ~soon. But maybe add a special flag for that?

Kixunil (commented):
While this doesn't resolve #14 completely, it seems to solve the most important parts. Channel opening could be added later if desired.

lnurl-withdraw.md
fiatjaf: I'm not sure about that very specific use case, but I think it makes sense to allow zero, specially considering automatic `balanceCheck`s.

lnurl-withdraw.md
```diff
@@ -77,8 +77,8 @@ Eg:
 	https://LNserviceURL
 	?tag=withdrawRequest
 	&k1=String
-	&minWithdrawable=MilliSatoshi
-	&maxWithdrawable=MilliSatoshi
+	&minWithdrawable=Integer
+	&maxWithdrawable=Integer
```
Kixunil, line 81: I personally find specifying units much easier to understand. What is the reason for this change?

lnurl-withdraw.md
```diff
@@ -17,8 +17,8 @@ Today users are asked to provide a withdrawal Lightning invoice to a service, th
         callback: String, // the URL which LN SERVICE would accept a withdrawal Lightning invoice as query parameter
         k1: String, // random or non-random string to identify the user's LN WALLET when using the callback URL
         defaultDescription: String, // A default withdrawal invoice description
-        maxWithdrawable: MilliSatoshi, // max withdrawable amount for a given user on LN SERVICE
-        minWithdrawable: MilliSatoshi, // An optional field, defaults to 1 MilliSatoshi if not present, can not be less than 1 or more than `maxWithdrawable`
+        minWithdrawable: Integer, // Min amount (in millisatoshis) the user can withdraw from LN SERVICE, or 0
```
Kixunil, line 20: Nit: "Min amount or 0" sounds a bit funny to me, I'd word it as "Min amount, can be 0"

lnurl-withdraw.md
chill117: I agree. Especially since the lnurl-channel spec uses whole sats instead of msats.

fiatjaf: I just find it clearer this way because "MilliSatoshi" is not a valid JSON type. But I can change this back.

fiatjaf:
Maybe a better and quicker explanation here: https://twitter.com/fiatjaf/status/1388216087817424899

PR #66, opened 2021-06-24
Change `lightning:` to `lnurl:`

### LINKED ISSUE #53: Change lightning: to lnurl: for App URL Schemes
By recommending that `lightning:` is used to open LNURLs for apps, it causes a clash with apps that don't or never will support LNURL, but use the `lightning:` scheme.

For example, Bottle Pay probably won't support LNURL (at least initially). And on iOS, if Bottle Pay is the last app that is installed with the `lightning:` scheme, any web pages or apps on the device will redirect to Bottle Pay by default, with no option to change the default behaviour in the operating system.

On Android the user can change default behaviour, but its buried in the OS settings and is a customer support issue that could be avoided.

Perhaps the scheme should be `lnurl:` to avoid this.

Fittiboy:
This changes the few occurrences of `lightning:` to `lnurl:`, which would close #53.
I can't currently think of a situation in which this would break existing implementations.

This simply allows wallets that implement this change first to override other wallets' behavior:
* On Android, it enables further customization of url scheme handling.
* On iOS, this makes the [url scheme lottery](https://developer.apple.com/documentation/xcode/defining-a-custom-url-scheme-for-your-app/#Register-Your-URL-Scheme)¹ a little bit less of a nightmare, as new `lnurl:` links will never be opened by wallets that don't support lnurl. I've looked into this and there seems to be literally no way to change url scheme behavior on iOS.

Potential future problem this can (and likely will) cause:
* Wallet developers lagging behind, and users of older wallets not being able to scan QR codes with the new `lnurl:` scheme

I would personally argue that this is a much better problem to have than wallets without lnurl support opening `lightning:LNURL...` links.

¹The only way to force a certain scheme to be opened by a _specific_ app is to register a reverse DNS string with the app, like `com.example.subdomain`, so that links like `scheme:subdomain.example.com` are opened by that app (from my understanding). Otherwise, it seems that currently it just opens the link with the latest app that has been installed (and registered support for that scheme).

### DIFF
--- a/lnurl-channel.md
+++ b/lnurl-channel.md
@@ -7,7 +7,7 @@ Suppose user has a balance on a certain service which he wishes to turn into an 
 ### Wallet to service interaction flow:
 ![Diagram showing interaction](media/diagrams/lnurl-channel-incoming-1.0.svg "Diagram showing interaction")
 
-1. User scans a LNURL QR code or accesses an `lightning:LNURL..` link with `LN WALLET` and `LN WALLET` decodes LNURL.
+1. User scans a LNURL QR code or accesses an `lnurl:LNURL..` link with `LN WALLET` and `LN WALLET` decodes LNURL.
 2. `LN WALLET` makes a GET request to `LN SERVICE` using the decoded LNURL.
 3. `LN WALLET` gets JSON response from `LN SERVICE` of form:
 
@@ -38,7 +38,7 @@ Suppose user has a balance on a certain service which he wishes to turn into an 
 ## Wallet to service interaction flow:
 ![Diagram showing interaction](media/diagrams/lnurl-channel-incoming-1.0.svg "Diagram showing interaction")
 
-1. User scans a LNURL QR code or accesses an `lightning:LNURL..` link with `LN WALLET` and `LN WALLET` decodes LNURL.
+1. User scans a LNURL QR code or accesses an `lnurl:LNURL..` link with `LN WALLET` and `LN WALLET` decodes LNURL.
 2. `LN WALLET` makes a GET request to `LN SERVICE` using the decoded LNURL.
 3. `LN WALLET` gets JSON response from `LN SERVICE` of form:
 
--- a/lnurl-pay.md
+++ b/lnurl-pay.md
@@ -5,7 +5,7 @@
 ### Wallet to service interaction flow:
 ![Diagram showing interaction](media/diagrams/lnurl-pay-1.0.svg "Diagram showing interaction")
 
-1. User scans a LNURL QR code or accesses an `lightning:LNURL..` link with `LN WALLET` and `LN WALLET` decodes LNURL.
+1. User scans a LNURL QR code or accesses an `lnurl:LNURL..` link with `LN WALLET` and `LN WALLET` decodes LNURL.
 2. `LN WALLET` makes a GET request to `LN SERVICE` using the decoded LNURL.
 3. `LN WALLET` gets JSON response from `LN SERVICE` of form:
 
--- a/lnurl-withdraw.md
+++ b/lnurl-withdraw.md
@@ -7,7 +7,7 @@ Today users are asked to provide a withdrawal Lightning invoice to a service, th
 ### Wallet to service interaction flow:
 ![Diagram showing interaction](media/diagrams/lnurl-withdraw-1.0.svg "Diagram showing interaction")
 
-1. User scans a LNURL QR code or accesses an `lightning:LNURL..` link with `LN WALLET` and `LN WALLET` decodes LNURL.
+1. User scans a LNURL QR code or accesses an `lnurl:LNURL..` link with `LN WALLET` and `LN WALLET` decodes LNURL.
 
 2. `LN WALLET` makes a GET request to `LN SERVICE` using the decoded LNURL.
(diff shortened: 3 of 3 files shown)


fiatjaf:
I think this breaks all the wallets. Only BlueWallet is looking for `lnurl:` for now, I think. All `lightning:lnurl...` links will stop working.

And then if the wallets change they will lose compatibility with apps that are using `lightning:` because they will start looking only for `lnurl:`.

What am I missing?

Fittiboy:
What I meant is that it won‘t break any existing links that still have the `lightning:` scheme, as all current wallets still handle it, and will always be able to handle it without changes.
I am of the opinion that it is best to separate `lnurl:` from `lightning:`, as it is not part of BOLT, and would always lead to problems with wallets that don‘t inplement lnurl.
This change would allow new `lnurl:` links to be handled only by wallets that actually support lnurl, while old links can still be handled by all wallets.
Only the transition period could cause problems, but this can be avoided by simply waiting for most wallets to implement `lnurl:` scheme support, before deploying any `lnurl:` links. It could become a little messy, but it is in the hands of the end user to decide whether or or not they want to switch to `lnurl:` at any given time.
Once most wallets support `lnurl:`, lnurl server implementations can simply switch to the new scheme as the default, avoiding compatibility issues.

Fittiboy:
I also see this as a precedent for future protocols. We don't know what else will be built, and I don't think it's a good idea to make it standard practice to use the `lightning:` url scheme for protocols that aren't part of BOLT. Just as `lightning:` is a separation from the `bitcoin:` url scheme.
Generally, I assume most lightning wallets will have lnurl functionality, but the switch to `lnurl:` allows developers to build minimalist wallets, or to test their wallets _before_ implementing lnurl, while lettings other wallets handle `lnurl:` links.

fiatjaf:
What I'm trying to say is that it doesn't make sense to just change the spec and wait until things are fixed. Probably no one from the current implementing wallets is reading this spec or following this thread, so they won't immediately react as we change. And then someone could come to read the spec because they want to implement it in their app, then they will use `lnurl:` and no wallet will work with their app.

That's why I'm saying we must ensure first that all wallets have implemented support for both schemes, then we can change the spec.

fiatjaf:
But since this is a hard fork anyway, why not seize the opportunity to improve other issues that would also be hard forks?

For example, removing the bech32 dependency would be a good thing, specially since it was established that [it doesn't provide the QR compression that someone thought it would](https://github.com/fiatjaf/lnurl-rfc/issues/15) and its only benefit is providing an `lnurl` prefix to identify the codes -- which can be made obsolete by the `lnurl:` scheme. This would turn LNURLs into actual URLs like `lnurl:domain.com/path?querystring`.

But then many decades ago someone suggested that the prefixes actually indicated what kind of LNURL was each. Although I liked that suggestion, it was dismissed because, again, it would require all wallets to change and didn't seem worth doing, but now could be the time, so we can have something like `lnurlp:domain.com/path`, `lnurlw:domain.com/path`, `lnurlc:domain.com/path`.

Fittiboy:
I actually like that idea a lot. I think it's worth compiling a list of these changes and work on making them all happen. I especially like the idea of removing the bech32 encoding!
A full list of major changes might be easier to vote for as well, when most of the changes were previously dismissed due to the hard fork requirements.

pseudozach:
NACK
I don't agree with the justification. 
You can't really avoid some wallets not supporting lnurl and opening lnurl links. Same thing is going to happen when a wallet supports lnurlp but scans lnurlw. Wallets should just handle it with a not supported error.
On the other hand services now need to change everything. Wallets need to announce they support multiple schemes.
Similarly with lnurlp: lnurlw: suggestion it's a lot of overhead for wallets & services to avoid a small number of errors.
If it's going to happen, we should have some versioning because I imagine migration will be a mess.

Fittiboy:
I understand your reasoning, but I think it‘s important to consider the fact that there will be more protocols built on top of lightning than just lnurl. Support for all these protocols across different wallets will be all over the place, so using the `lightning:` url scheme for all of them would lead to disaster.
Lnurl isn‘t special, but simply the first major protocol on top of lightning. From this viewpoint, using `lightning:` makes little sense to begin with, just like using`bitcoin:` would be an odd choice.
For any protocol (this includes each lnurl sub-protocol), the default assumption should be that any given wallet does not support it, and that it should be opened only by wallets that explicitly implement it.
Migration might be somewhat messy, but right now it‘s still more than feasible to directly coordinate with wallet developers who already support the current version or lnurl.

fiatjaf:
@pseudozach 

> Same thing is going to happen when a wallet supports lnurlp but scans lnurlw.

Different schemes for each protocol fix that.

> Wallets need to announce they support multiple schemes.

What is this "announce" thing? Isn't it just writing a string on a list or something? It makes sense since they have already WRITE THE CODE for each separate scheme separately.

I don't see how adding a single letter is an overhead.

Versioning doesn't help at all with migration. Bumping a version is just a way to say you're making everything incompatible and pushing the burden on others.

Fittiboy:
Closing because of LUD-17

Fittiboy:
https://github.com/fiatjaf/lnurl-rfc/blob/luds/17.md

PR #67, opened 2021-07-03
Extract payment links from internet identifiers

btcontract:
Also added a `proofofpayer` query parameter in response call. 

It may come especially handy when paying to links extracted from emails since these pay links are very likely to be non-personalized (anyone can obtain lnurl-pay from email as opposed to service providing a personalized pay link to a logged in user).


btcontract:
Changed `email` to `internet identifier` everywhere to make it more generic. Also there are now two new metadata records: `text/email` or `text/identifier` to allow wallets to display these links as actual emails or as some generic identifiers, this will give users correct expectations right away.

PR #68, opened 2021-07-06
Change hardcoded endpoint

btcontract:


takinbo:
There's another section you'll need to make modifications:
```
an example for `payments@site.com` is GET `https://site.com/lnurl-pay?key=payments`
```
should be changed to:
```
an example for `payments@site.com` is GET `https://site.com/.well-known/lnurlp/payments`
```

PR #71, opened 2021-07-16
lnurl-withdraw - emphasise LN service must verify k1

pvyhnal-generalbytes:


pvyhnal-generalbytes:
note this page for example does not mention the `k1` in the second step at all and they probably use their own parameter `ott` instead.

https://docs.lnpay.co/lnurl-withdraw/how-does-lnurl-withdraw-work

fiatjaf:
Why is that an issue? k1 should not even be in this spec actually. Let people evaluate if they want to use it or rely on other params.

(But you must keep including it nonetheless since some wallets or services may expect it.)

pvyhnal-generalbytes:
ok, in that case the that should be mentioned there that you can use any other parameter instead of k1, right now it also says:
> Note that service will withdraw funds to anyone who can provide a valid ephemeral k1.

Maybe change this sentence so it's clear that the service needs to use some parameter to validate the request? Just because it's important for security when implementing a LN-SERVICE

hsjoberg:

PR #73, opened 2021-07-25
Update lnurl-withdraw response code suggestions.

microlancer:
Added tips/suggestions for handling various HTTP status codes returned from balanceCheck for smoother operation by following usual HTTP semantics.


hsjoberg:
The LNURL protocol is explicitly says that implementors should not assume HTTP codes have any meaning.
Anyway I'm closing this as we've moved to the LUD system. A new pull request can be opened.

PR #74, opened 2021-07-26
Allow self-signed certificates for onion addresses

btcontract:
Previously a plain HTTP was allowed for onion addresses but it turned out that not all clients are happy with this (Android phones prohibit all HTTP traffic for example). As a fix, we now require HTTPS for onion links as well, but also allow self-signed certificates for them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,9 +21,9 @@ Once `LNURL` is decoded:
 - If `tag` query parameter is present then this `LNURL` has a special meaning, further actions will be based on `tag` parameter value.
 - Otherwise a GET request should be executed which must return a JSON object containing a `tag` field, further actions will be based on `tag` field value.
 
-## HTTPS or Onion
+## Self-signed certificates for Onion addresses
 
-`LNURL` is acceptable in two forms: either an `https://` clearnet link (no self-signed certificates allowed) or an `http://` v2/v3 onion link.
+`LNURL` links must use `https` at all times, self-signed certificates are only allowed for v2/v3 onion addresses.
 
 ## Fallback scheme
 
--- a/lnurl-pay.md
+++ b/lnurl-pay.md
@@ -10,7 +10,7 @@
 1.1.2. `LN WALLET` makes a GET request to `LN SERVICE` using the decoded LNURL.  
 **Or**   
 1.2.1. User scans/pastes/shares an [internet identifier](https://datatracker.ietf.org/doc/html/rfc5322#section-3.4.1) with `LN WALLET`.  
-1.2.2. `LN WALLET` extracts `domain` and `user` parts from identifier, makes a GET request to `https://<LN SERVICE domain>/.well-known/lnurlp/<user>` endpoint if `domain` is clearnet or `http://<LN SERVICE domain>/.well-known/lnurlp/<user>` if `domain` is onion, an example for `payments@site.com` is GET `https://site.com/.well-known/lnurlp/payments`.  
+1.2.2. `LN WALLET` extracts `domain` and `user` parts from identifier, makes a GET request to `https://<LN SERVICE domain>/.well-known/lnurlp/<user>` endpoint, an example for `payments@site.com` is GET `https://site.com/.well-known/lnurlp/payments`.  
 **Then**  
 2. `LN WALLET` gets JSON response from `LN SERVICE` of form:
 


hsjoberg:
ACK

btcontract:
Unless there are substantial arguments against this I'd suggest we move on with merging as users start having issues with current scheme.

theborakompanioni:
I haven't had the time to try lnurl-auth with HTTP over TOR on Android. Generally the state of HTTPS on Tor is still quite a horror. I would prefer if plain HTTP over TOR could stay in spec. Does anyone have experience with Tor on Android and what the issues are exactly?

fiatjaf:
I've presented my substantial arguments on https://t.me/lnurl. They remain unaddressed.

btcontract:
>Does anyone have experience with Tor on Android and what the issues are exactly

The issue is that all kinds of http traffic are prohibited on Android and there is no clear fix: there is a manifest directive which allows an app to use http, but it works since API level 26 which means it won't work for substantial amount of users + requiring an app to allow http traffic in as security-sensitive app as bitcoin wallet will definitely prove to be controversial.

@theborakompanioni, taken from your reply in https://github.com/btcontract/lnwallet/issues/39
> Self-signed certificates over tor will also display a cert warning in tor browser.

Do I understand it correct that Tor browser has no issues with `http` but will display warnings for self-signed `https`? If so, is this a viable setup:

1. Use plain `http` on `<key>.onion:80` endpoint, the one user visits through Tor browser.
2. User self-signed `https` on `<key>.onion:443` or `auth.<key>.onion` subdomain specifically tailored for lnurl-auth.

If such combination is possible we can make both Tor browser and Android devices happy.

btcontract:
@fiatjaf 
> I've presented my substantial arguments on https://t.me/lnurl. They remain unaddressed.

They were addressed by me, copied once again in https://github.com/fiatjaf/lnurl-rfc/pull/74#issuecomment-890171861

theborakompanioni:
> The issue is that all kinds of http traffic are prohibited on Android and there is no clear fix: there is a manifest directive which allows an app to use http, but it works since API level 26 which means it won't work for substantial amount of users + requiring an app to allow http traffic in as security-sensitive app as bitcoin wallet will definitely prove to be controversial.

Yes. I am with you. From this point of view it seems going for HTTPS-only, also in the spec, is the better choice.

> Do I understand it correct that Tor browser has no issues with `http` but will display warnings for self-signed `https`? If so, is this a viable setup:
> 
> 1. Use plain `http` on `<key>.onion:80` endpoint, the one user visits through Tor browser.
> 2. Use self-signed `https` on `<key>.onion:443` or `auth.<key>.onion` subdomain specifically tailored for lnurl-auth.
>
> If such combination is possible we can make both Tor browser and Android devices happy.

I cannot believe I did not think of that. This will work.
Thank you.

fiatjaf:
My points are on https://t.me/lnurl/12813

image

It basically unmounts to being unclear to me if this is a desirable solution for everybody, or if you're imposing a burden on everybody else and the protocol because of an Android bug. I think we must wait for more input from people who actually use Tor and specially onion services that are in process of implementing this just so we don't rush any stupid decisions.

I also don't understand why it "weakens security" to allow HTTP calls, it really doesn't make any sense to me. Are you saying browsers have very weak security, for example? They do allow HTTP. And `bitcoin-cli`, for example, is an app that uses HTTP to talk to `bitcoind`. Is that essentially unsafe to even _allow_ HTTP? I think that depends on the context and I don't see that being the case for a Bitcoin wallet.

theborakompanioni:
I guess right now we can settle on that it should at least be possible to use https with tor. Right?
Maybe lets just allow self-signed certs for https on onion addresses for now. Optionally add a disclaimer like `be aware that some mobile operating systems do not allow calls to http-only onion address` (or something like that).

btcontract:
@fiatjaf allowing `http` weakens security relative to Android platform defaults. Even if we forget about this aspect for a moment there is no clear way to allow `http` either, pre API 26 users will still have this issue (I think  I'm repeating this for the third time now).

IMO this is the kind of thing which should be applied ASAP because it's the only clear way to do stuff now that we know about Android specifics.

btcontract:
Resolved this by making a security exception as follows: https://github.com/btcontract/lnwallet/commit/73b92274955b132da655fcf1072066bd34cfdae5.

This is still not ideal but better security wise since only onion domains are affected and it works for API24+.

PR #82, opened 2021-08-17
Fix Blixt Wallet not supporting LUD-05 but supports LUD-13

hsjoberg:
Blixt Wallet uses the "node message signing" way of deriving linkingKey
in the auth protocol.


fiatjaf:
Addressed in dfdba78dc9e028c1aed4842ed5ef6db958ae84ab.

PR #84, opened 2021-08-18
Describe in LUD-13's title that it's about the common "signMessage" API

hsjoberg:
While the purpose of LUD-13 was to have a standard way for lnd-based wallets to implement LNURL-auth, Zeus is now using the node message signature method for generating the hashingKey for all their backends (lnd, c-lightning and eclair).

### DIFF
--- a/13.md
+++ b/13.md
@@ -1,4 +1,4 @@
-LUD-13: LND-based seed generation for auth protocol.
+LUD-13: `signMessage`-based seed generation for auth protocol.
 ====================================================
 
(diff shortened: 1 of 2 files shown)


fiatjaf:
We could say "`signMessage`-based seed generation for auth protocol.", because they're all using this `signMessage` API which is its own protocol. It's not just the node signing, it is actually the node signing using a special HMAC with its own standard phrase and algorithm.

hsjoberg:
@fiatjaf Okay, that makes sense and goes along better with the actual content of the LUD.
I've updated the PR.

PR #89, opened 2021-09-02
adding BTCPayServer

Kukks:


fiatjaf:
Why is this a work in progress? Can we merge now and change later if necessary?

hsjoberg:
Does BTCPayServer support [LUD-12 Comments in payRequest](https://github.com/fiatjaf/lnurl-rfc/blob/luds/12.md) and [LUD-16 Paying to static internet identifiers](https://github.com/fiatjaf/lnurl-rfc/blob/luds/16.md)?

PR #90, opened 2021-09-05
Change proof of payer to payer identity

btcontract:
This revamps LUD-18 in the following ways:

- It renames "proof of payment" to "payer identity" thus emphasizing there are no strong cryptographic proofs but instead a rather loose identity data which is mostly intended for accounting purposes on `LN SERVICE` side.
- It allows many types of identities to be provided building on current LNURL subprotocols.


18.md
```diff
@@ -1,19 +1,46 @@
...
-
-The `proofofpayer` is an optional ephemeral secp256k1 public key generated by payer, a corresponding private key should be retained by payer, a payee may later ask payer to provide a public key itself or sign a random message using corresponding private key and thus provide a proof of payer.
-
-The callback is thus modified:
+Payer MAY attach a `payerid` query parameter to LNURL-PAY callback with value set to `base64`-encoded JSON array:
 
 ```diff
 - <callback><?|&>amount=<milliSatoshi>
-+ <callback><?|&>amount=<milliSatoshi>&proofofpayer=<hex(ephemeral secp256k1 public key bytes)>
++ <callback><?|&>amount=<milliSatoshi>&payerid=<base64(<array of identity objects>)>
```
hsjoberg, line 14: I suggest relying on "url encoding" here would be better than base64'ing the string, which would also need encoding because of the `=` symbol being a part of the base64 charset.

```JS
encodeURIComponent(<JSON string>)
```

btcontract:
Base64URL:

`[{tag: "text/plain", sender: "mr. X"}, {tag: "text/identifier", sender: "user@identifier.com"}]` = `W3t0YWc6ICJ0ZXh0L3BsYWluIiwgc2VuZGVyOiAibXIuIFgifSwge3RhZzogInRleHQvaWRlbnRpZmllciIsIHNlbmRlcjogInVzZXJAaWRlbnRpZmllci5jb20ifV0`


URL encoding: 

`[{tag: "text/plain", sender: "mr. X"}, {tag: "text/identifier", sender: "user@identifier.com"}]` = `%5B%7Btag%3A%20%22text%2Fplain%22%2C%20sender%3A%20%22mr.%20X%22%7D%2C%20%7Btag%3A%20%22text%2Fidentifier%22%2C%20sender%3A%20%22user%40identifier.com%22%7D%5D`

hsjoberg:
Interesting, with just url encoding there are too many symbols that have to be encoded in a JSON string.

I'm in favor of a normal base64 being url encoded then. The symbol in base64 that needs encoding is `=`.
Having the implementer import another lib just to be able to use this LUD seems off.

This can trivially be solved by `urlencode(base64(<json string>))` and `base64` and `urlencode` should be widely available, often in the standard lib (`urlencode` is already needed for comment (LUD-12).

URL encoding over Base64 encoding: 

`[{tag: "text/plain", sender: "mr. X"}, {tag: "text/identifier", sender: "user@identifier.com"}]` = 

`W3t0YWc6ICJ0ZXh0L3BsYWluIiwgc2VuZGVyOiAibXIuIFgifSwge3RhZzogInRleHQvaWRlbnRpZmllciIsIHNlbmRlcjogInVzZXJAaWRlbnRpZmllci5jb20ifV0%3D`

(`%3D` is the difference compared to Base64URL and there can only be this much of a difference because of how padding works in Base64. There can only be one `=`).

hsjoberg:
Although I'm not sure the string difference is not that much longer to make this worthwhile, especially considering the added complexity.

fiatjaf:
Most or all http servers and clients will already urlencode and urldecode things automatically if we just pass ugly JSON blobs to them. Is size really a concern here? Why? I don't think it is.

hsjoberg (approved):
LGTM

hsjoberg:
Currently there is no way for the service to tell the wallet what kind of payerid metadata it expects from the wallet.
The wallet has to either make a guess or send everything it has available.

 
The initial LNURL response object could include a field `acceptsPayerId` that tells the wallet and thus the user what kind of ids to embed.

```
SERVICE MAY include require or request payer id's by including in the LURL payload.
{
    callback: String,
    maxSendable: MilliSatoshi,
    minSendable: MilliSatoshi,
    metadata: String,
    tag: "payRequest"
    acceptsPayerIds: [<[identity record, mandatory]> ...]
}
```

When ordering in an online shop `acceptsPayerIds` could look like this:


```
acceptsPayerIds: [["application/lnurl-auth, true]]
```


When sending a payment to a Lightning Address, it could look like this

```
acceptsPayerIds: [["text/identifier, false]]
```

The wallet MUST display what metadata will be sent to the service upon paying.

hsjoberg:
Alternative approach that puts it in the LNURL-pay `metadata` payload and thus verifiable:

> SERVICE MUST request payer id's by including a `application/payer-ids` entry in the LNURL-pay `metadata` payload:
> 
> ```
> [
>   ...,
>   [
>     "application/payer-ids", 
>     ["text/plain"],
>     ...,
>   ],
> ]
> ```
>
> Upon recognizing payer id's WALLET SHOULD asks the user whether to include them.
> WALLET SHOULD NOT include any payer id's not requested by the SERVICE.
> 
> ## `LNURL-AUTH` integration specific
> 
> ...

hsjoberg (approved):

PR #94, opened 2021-09-14
lud-17 fallback scheme.

fiatjaf:


17.md
```diff
@@ -26,6 +26,16 @@ This document defines that LNURLs will have the format of conventional URLs, wit
...
 
+## Fallback scheme
+
+To replace the [_Fallback scheme_ from LUD-01](01.md#fallback-scheme), wallets should also allow for `https://`-prefixed URLs when scanning QR codes and handle them just like they handle `lnurl1...` links today.
+
+The idea is that the fallback scheme is expected to only be used in QR codes, never in links (the original intent of the fallback scheme was to be used in printed lnurl-withdraw vouchers, but this also applies to printed lnurl-pay QR codes like the [McDonald's ones](https://twitter.com/fiatjaf/status/1435251011527000066)).
+
+If someone sees a QR code printed on a place, for example, they may know it's an LNURL thing and scan directly with their Lightning wallet, but they may also think it is a normal QR code or they don't even have a Lightning wallet, so they scan it and try to open it on their web browser. In that case the `SERVICE` may handle the request, check for the presence of `Accept: text/html` header and return a webpage with instructions.
+
+This remains entirely optional and `SERVICE` should only use `https://` in printed QR codes if it is also willing to return a fallback HTML page.
```
chill117, line 37: Does this advice refer only to the case where the URL is encoded directly as a QR code and **not** when encoded using bech32 (e.g. "lnurl1...") then encoded as QR code?

fiatjaf: Yes.

augustresende (approved):

augustresende:
@fiatjaf Maybe we can improve this fallback for QR Codes keeping LUD-1 fallback scheme

like this

https://someendpoint.com/lnurl/123456?lightning=LNURL1DP68GURN8GHJ7UM0D4JK2MNYWPHKJMN59E3K7MF0D3H82UNV9UCNYVE5X5MQEQFRPL < (that is this same previous link)

This string encoded as QR code can work with old (only LUD-1 without change) and new (LUD-17) implementations without any problem

fiatjaf:
I like this. @AugustoResende's suggestions covers all cases.

Services can choose to either do the `Accept` header magic or append the LUD-01 LNURL to the https:// URL -- or both.

augustresende:
When we can merge this? @fiatjaf @hsjoberg

dni:
bump

dni (approved):

PR #96, opened 2021-09-24
LUD-18 as a pay-back scheme

rafaelpac:
I believe that LNURL-pay should provide a way for the service to pay back some change to the user.
The use case is for stuff that the price can't be known beforehand, like a gas pump, rental equipment...
The user pre-pay what should be a reasonable amount, a little bit more than average, and receive a change back at the end.
Somewhat related to what I commented here:
https://github.com/fiatjaf/lnurl-rfc/pull/49#issuecomment-826336746
Therefore, this LUD-18 come in handy, as the user identifies himself in a useful way, providing a way to receive the change.


fiatjaf:
Yes, that's one of the use cases for LUD-18. However for now that flow will have to be fully manual.

dni:
i really like this @hsjoberg :)

andrerfneves:
Closing this as it's super stale. We can always reopen.

PR #100, opened 2021-09-30
fix LUD-18 because it was breaking implementations.

fiatjaf:


18.md
```diff
@@ -16,41 +16,34 @@ Payer MAY attach a `payerid` query parameter to LNURL-PAY callback with value se
...
++  payerIDs: {
++    "name": Any,
++    "pubkey": Any,
++    "identifier": Any,
++    "email": Any,
++    "auth": {
++       "mandatory": Boolean,
++       "k1": String // hex encoded 32 bytes of challenge
++    ]
++  },
```
dpad85, line 44: How about:

```
payerIDs: {
  "mandatory": JSONArray<String>,
  "optional": JSONArray<String>
}
```

Where the strings in those json arrays could be `name`, `pubkey`, `identifier`, `email`, `auth:random_hex`?

If `Wallet` does not support a mandatory field (e.g `email`), it would reject the lnurl.

fiatjaf: Good point, indeed. That is flexible enough for other kinds of payerids to be added in the future. Like `home_address` or something like that.

fiatjaf: But I don't like the `auth:random_hex` part, I'm afraid of string parsing. What about

```
   payerIDs: {
     "name": { "mandatory": Boolean },
     "pubkey": { "mandatory": Boolean },
     "identifier": { "mandatory": Boolean },
     "email": { "mandatory": Boolean },
     "auth": {
        "mandatory": Boolean,
        "k1": String // hex encoded 32 bytes of challenge
     ]
   },
```

dpad85: LGTM 👍

fiatjaf:
I was confused and made two PRs.

Check https://github.com/fiatjaf/lnurl-rfc/pull/101 please.

PR #101, opened 2021-09-30
fix LUD-18 because it was breaking implementations.

fiatjaf:
And also make it so payerids are committed to the invoice.

### DIFF
--- a/18.md
+++ b/18.md
@@ -69,7 +69,7 @@ On its own side, `WALLET` must do the same thing before hashing the metadata and
 
 If `WALLET` sends more `payerid` fields than requested by `SERVICE`, `SERVICE` should still append everything as received.
 
-`WALLET` must not send expected `SERVICE` to append to metadata if `SERVICE` hadn't included the `payerIDs` record in the first response, as that would mean `SERVICE` doesn't understand LUD-18 at all.
+`WALLET` must not expect `SERVICE` to append the identity records to the metadata if `SERVICE` hadn't included the `payerIDs` record in the first response, as that would mean `SERVICE` doesn't understand LUD-18 at all.
 
 ### Pseudocode example for calculating `descriptionHash` with `payerids`
 
--- a/18.md
+++ b/18.md
@@ -54,16 +54,17 @@ The mere presence of each key in the `payerIDs` record signals that it CAN be in
 
 If `SERVICE` requests (section 2) and `WALLET` sends a `payerid` record in the callback (section 1), the payer IDs must be committed to the metadata before the invoice is created.
 
-`SERVICE` MUST append it to the metadata as it was received, inside a new record called `application/payerids`, as follows:
+`SERVICE` MUST append it to the metadata as it was received (after url-decoding), as in the following example:
 
 ```diff
- [
-   ["text/plain", "description"], // example
-   ["image/png;base64", "AAA=="], // example
-+  ["application/payerids", <array of identity records as received from WALLET>]
- ]
+ [["text/plain", "description"], ["image/png;base64", "AAA=="]]
++[["text/plain", "description"], ["image/png;base64", "AAA=="]][<identity record>, ...]
 ```
 
 After doing that, `SERVICE` can proceed to hash the metadata and include that hash in the generated BOLT11 invoice as the `descriptionHash` field.
 
 On its own side, `WALLET` must do the same thing before hashing the metadata and checking its hashed value against the `descriptionHash` field of the invoice received from `SERVICE`.
+
+If `WALLET` sends more `payerid` fields than requested by `SERVICE`, `SERVICE` should still append everything as received.
+
+`WALLET` must not send expected `SERVICE` to append to metadata if `SERVICE` hadn't included the `payerIDs` record in the first response, as that would mean `SERVICE` doesn't understand LUD-18 at all.
--- a/18.md
+++ b/18.md
@@ -1,7 +1,7 @@
 LUD-18: Payer identity in `payRequest` protocol.
 ================================================
 
-`author: akumaigorodski`
+`author: akumaigorodski` `author: hampus_s` `author: fiatjaf`
 
 ---
 
@@ -16,41 +16,34 @@ Payer MAY attach a `payerid` query parameter to LNURL-PAY callback with value se
 
 JSON array may contain one or more records of the following forms:
 
-1. `[ "text/plain", <free-form string> ]`
-2. `[ "application/pubkey", hex(<randomly generated secp256k1 pubkey>) ]`
-3. `[ "application/lnurl-auth", hex(<linkingKey>), hex(sign(utf8ToBytes(<k1>), <linkingPrivKey>)) ]`
-4. `[ "text/identifier", <internet identifier> ]`
-5. `[ "text/email", <email address> ]`
+1. `{ "tag": "name", "value": <free-form string> }`
+2. `{ "tag": "pubkey", "value": hex(<randomly generated secp256k1 pubkey>) }`
+3. `{ "tag": "auth", "value": hex(<linkingKey>), "sig": hex(sign(utf8ToBytes(<k1>), <linkingPrivKey>)) }`
+4. `{ "tag": "identifier", "value": <internet identifier> }`
+5. `{ "tag": "email", "value": <email address> }`
 
-If present, JSON array MAY NOT contain more than one record with the same metadata type.
+## `PayerIDs` record
 
-## Payer identity metadata record
+If `SERVICE` wants to get one or more types of payer identities from `WALLET` then it must alter its JSON response to the first callback to include a `payerIDs` field, as follows:
 
-If `LN SERVICE` wants to get one or more types of payer identities from `LN WALLET` then it MUST put the following record into metadata which contains `application/payer-ids` string as first item further followed by records each indicating a type of identitiy that `LN WALLET` is allowed to provide. 
-
+```diff
+ {
+   callback: String,
+   maxSendable: MilliSatoshi,
+   minSendable: MilliSatoshi,
+   metadata: String,
++  payerIDs: {
++    "name": Any,
++    "pubkey": Any,
++    "identifier": Any,
++    "email": Any,
++    "auth": {
++       "mandatory": Boolean,
++       "k1": String // hex encoded 32 bytes of challenge
++    ]
++  },
+   tag: "payRequest",
+ }
 ```
-[
-  ...,
-  [
-    "application/payer-ids", 
-    [ "text/plain" ],
-    [ "application/pubkey" ],
-    [ 
-      "application/lnurl-auth", 
-      isMandatory, // boolean which flags if user MAY include `application/lnurl-auth` record in callback or MUST do that
-      k1 // hex encoded 32 bytes of challenge
-    ],
-    [ "text/identifier" ],
-    [ "text/email" ],
-    ...,
-  ],
-]
-```
 
-`LN WALLET` SHOULD NOT send payer identity types omitted in `application/payer-ids` metadata record, none at all if record is not present.
-
-Upon seeing an `application/lnurl-auth` record with `isMandatory` set to `false` user SHOULD be explicitly given an option to include an `LNURL-AUTH` record in callback.
-
-Upon seeing an `application/lnurl-auth` record with `isMandatory` set to `true` user MUST be informed by `LN WALLET` that callback will contain an `LNURL-AUTH` record should a payment be sent.
-
-When including an `LNURL-AUTH` record in a callback a `linkingKey` MUST be the one originally seen in `LNURL-PAY` static link.
+The mere presence of each key in the `payerIDs` record signals that it CAN be included by `WALLET`. The values don't mean anything and are to be discarded -- with the exception of `auth`, which MUST be an object containing `k1` and `mandatory`. In case `mandatory` is true `WALLET` MUST send the payment with a [lnurl-auth](04.md) key and signature and the tag `auth`.
--- a/18.md
+++ b/18.md
@@ -21,6 +21,7 @@ JSON array may contain one or more records of the following forms:
 3. `{ "tag": "auth", "value": hex(<linkingKey>), "sig": hex(sign(utf8ToBytes(<k1>), <linkingPrivKey>)) }` // following [LUD-04](04.md)
 4. `{ "tag": "identifier", "value": <internet identifier> }`
 5. `{ "tag": "email", "value": <email address> }`
+6. other fields can be negotiated.
 
 ## 2. `PayerIDs` record
 
@@ -40,7 +41,8 @@ If `SERVICE` wants to get one or more types of payer identities from `WALLET` th
 +    "auth": {
 +       "mandatory": Boolean,
 +       "k1": String // hex encoded 32 bytes of challenge
-+    ]
++    ],
++    ...other fields may be negotiated
 +  },
    tag: "payRequest",
  }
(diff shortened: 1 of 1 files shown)


18.md
```diff
@@ -1,11 +1,11 @@
 LUD-18: Payer identity in `payRequest` protocol.
 ================================================
 
-`author: akumaigorodski`
+`author: akumaigorodski` `author: hampus_s` `author: fiatjaf`
 
 ---
 
-## Specifying payer identity before sending a payment
+## 1. Specifying payer identity before sending a payment
```
dpad85: I think this section should be moved after `## 2. PayerIDs record`. It makes more sense that the specs first define the `SERVICE`'s request, then how the `WALLET` handles that (which is by adding a new `payerid` param to the callback).

Maybe a introduction could be added to give the rationale about all this (which would be, letting `SERVICE` attach a real-world id to the payer).

18.md
```diff
@@ -16,41 +16,66 @@ Payer MAY attach a `payerid` query parameter to LNURL-PAY callback with value se
...
-`LN WALLET` SHOULD NOT send payer identity types omitted in `application/payer-ids` metadata record, none at all if record is not present.
+After doing that, `SERVICE` can proceed to hash the metadata and include that hash in the generated BOLT11 invoice as the `descriptionHash` field.
+
+On its own side, `WALLET` must do the same thing before hashing the metadata and checking its hashed value against the `descriptionHash` field of the invoice received from `SERVICE`.
 
-Upon seeing an `application/lnurl-auth` record with `isMandatory` set to `false` user SHOULD be explicitly given an option to include an `LNURL-AUTH` record in callback.
+If `WALLET` sends more `payerid` fields than requested by `SERVICE`, `SERVICE` should still append everything as received.
 
-Upon seeing an `application/lnurl-auth` record with `isMandatory` set to `true` user MUST be informed by `LN WALLET` that callback will contain an `LNURL-AUTH` record should a payment be sent.
+`WALLET` must not send expected `SERVICE` to append to metadata if `SERVICE` hadn't included the `payerIDs` record in the first response, as that would mean `SERVICE` doesn't understand LUD-18 at all.
```
dpad85:
```suggestion
`WALLET` must not expect `SERVICE` to append the identity records to the metadata if `SERVICE` hadn't included the `payerIDs` record in the first response, as that would mean `SERVICE` doesn't understand LUD-18 at all.
```
Is that what you meant?

18.md
```diff
@@ -16,41 +16,66 @@ Payer MAY attach a `payerid` query parameter to LNURL-PAY callback with value se
...
-2. `[ "application/pubkey", hex(<randomly generated secp256k1 pubkey>) ]`
-3. `[ "application/lnurl-auth", hex(<linkingKey>), hex(sign(utf8ToBytes(<k1>), <linkingPrivKey>)) ]`
-4. `[ "text/identifier", <internet identifier> ]`
-5. `[ "text/email", <email address> ]`
+1. `{ "tag": "name", "value": <free-form string> }`
+2. `{ "tag": "pubkey", "value": hex(<randomly generated secp256k1 pubkey>) }`
+3. `{ "tag": "auth", "value": hex(<linkingKey>), "sig": hex(sign(utf8ToBytes(<k1>), <linkingPrivKey>)) }` // following [LUD-04](04.md)
+4. `{ "tag": "identifier", "value": <internet identifier> }`
+5. `{ "tag": "email", "value": <email address> }`
+6. other fields can be negotiated.
```
dpad85: To make things more consistent with the `payerIDs` field in the request sent by the service, how about having the `payerid` parameter like this?

```
[
  "name": <free form string>,
  "pubkey": hex(<randomly generated secp256k1 pubkey>),
  "auth": {
    "value": hex(<linkingKey>),
    "sig": hex(sign(utf8ToBytes(<k1>), <linkingPrivKey>)) // following LUD-04
  },
  "email": <email address>,
  "identifier": <internet identifier>
]
```

And I also think it should be stated that this json must be then stringified (maybe remove spaces/blank chars) and url encoded. And that this **exact value** will be used later when checking the `SERVICE`'s invoice. It's not really obvious at first.

dpad85 (commented):
Moving the payer id request away from the metadata is a good change IMO, because the `Service` is actually requesting a possibly mandatory data from the `Wallet`. In that sense it's similar to [LUD-12](https://github.com/fiatjaf/lnurl-rfc/blob/luds/12.md) that adds a `commentAllowed` field.

Since the payer id info are baked into the invoice's description hash, this will impact the check made by wallets in the basic lnurl-pay flow. So maybe we should mention that in LUD-06, so that they can fail this LNUrl gracefully?

18.md
hsjoberg: @dpad85 It doesn't have to be stringified though? Definitely url encoded, but that's expected anyway.

Agree on changing the format.

fiatjaf:
> Since the payer id info are baked into the invoice's description hash, this will impact the check made by wallets in the basic lnurl-pay flow. So maybe we should mention that in LUD-06, so that they can fail this LNUrl gracefully?

How would it impact the check from LUD-06? If it will, then we shouldn't change this here at all. The entire goal is to never have breaking changes.

My rationale for concluding this won't be a breaking change considers the following cases:

1. `SERVICE` and `WALLET` both not support LUD-18: all good.
2. `SERVICE` and `WALLET` both support LUD-18: all good.
3. `SERVICE` supports LUD-18, `WALLET` doesn't: `SERVICE` will ask for payerids, `WALLET` won't provide them, so there will be nothing to append to `metadata`, everything stays the same.
4. `SERVICE` does not support LUD-18, `WALLET` does: `SERVICE` won't ask for payerids, `WALLET` won't provide them, so again nothing to append.

18.md
```diff
@@ -1,56 +1,96 @@
...
+If `SERVICE` wants to get one or more types of payer identities from `WALLET` then it must alter its JSON response to the first callback to include a `payerIDs` field, as follows:
+
+```diff
+ {
+   callback: String,
+   maxSendable: MilliSatoshi,
+   minSendable: MilliSatoshi,
+   metadata: String,
++  payerIDs: {
++    "name": { "mandatory": Boolean },
```
Kukks, line 23: Are all of these really necessary to be in the payload? Seems excessive. `payerIDs` should be a freeform  specification requirement defined by `SERVICE`. with a schema:`{ [fieldName:string]: { mandatory?:boolean, format?:string}}`

Kukks (changes requested):
I dislike how much fluff this adds. I don't see why we MUST force a set of those fields to be included ALL the time. The query string to send from `WALLET` becomes huge ALL the time..

@btcontract has a good point that this LUD has the potential to replace some of the looser LUDs like comments and name in comment. https://t.me/lnurl/24533 
If there's consensus, this LUD would best find a way to make it backwards compatible with those LUDs.

akumaigorodski:
Why is there a mandatory flag everywhere again? This excludes a class of wallets who can't or don't want to deal with emails/ids. Mandatory is only justified for auth for its specific reasons.

18.md
hsjoberg: Everything isn't necessary in the payload. Note that we're not using Typescript definitions in LNURL.
This is just listing every possible payerid.

18.md
```diff
@@ -1,56 +1,96 @@
...
 
-## Specifying payer identity before sending a payment
+The idea here is that the payer can identify itself when paying. The scope varies from free-form names (or [LUD-16](16.md) internet identifiers) useful for loose identification in comments or similar things, to cryptographic keys for proof of payment, to authentication keys for future login into a `SERVICE`'s website after a payment and other use cases.
 
-Payer MAY attach a `payerid` query parameter to LNURL-PAY callback with value set to JSON array:
+Aside from that, the payer ids are also committed to the invoice by means of the `descriptionHash` property, which ensures strong cryptographic guarantees for proof of payment.
+
+## 1. `payerIDs` record
+
+If `SERVICE` wants to get one or more types of payer identities from `WALLET` then it must alter its JSON response to the first callback to include a `payerIDs` field, as follows:
```
hsjoberg:
```diff
--- If `SERVICE` wants to get one or more types of payer identities from `WALLET` then it must alter its JSON response to the first callback to include a `payerIDs` field, as follows:
+++ If `SERVICE` wants to get one or more types of payer identities from `WALLET` then it MUST alter its JSON response to the first callback to include a `payerIDs` field, as follows:

File: 18.md [hunk 6]
```diff
@@ -1,56 +1,96 @@
...
 
-JSON array may contain one or more records of the following forms:
+Each key in this JSON object should correspond to a requested payerid from the `payerIDs` record received from `SERVICE`.
 
-1. `[ "text/plain", <free-form string> ]`
-2. `[ "application/pubkey", hex(<randomly generated secp256k1 pubkey>) ]`
-3. `[ "application/lnurl-auth", hex(<linkingKey>), hex(sign(utf8ToBytes(<k1>), <linkingPrivKey>)) ]`
-4. `[ "text/identifier", <internet identifier> ]`
-5. `[ "text/email", <email address> ]`
+`WALLET` CAN send any of the payer id kinds if they are listed in the `payerIDs` record. But if any is marked as `"mandatory": true` `WALLET` must send it if it can, otherwise do not proceed with the payment flow.
```
hsjoberg:
```diff
--- `WALLET` CAN send any of the payer id kinds if they are listed in the `payerIDs` record. But if any is marked as `"mandatory": true` `WALLET` must send it if it can, otherwise do not proceed with the payment flow.
+++ `WALLET` CAN send any of the payer id kinds if they are listed in the `payerIDs` record. But if any is marked as `"mandatory": true` `WALLET` MUST send it if it can, otherwise do not proceed with the payment flow.

File: 18.md [hunk 7]
```diff
@@ -1,56 +1,96 @@
...
 
-If present, JSON array MAY NOT contain more than one record with the same metadata type.
+`WALLET` SHOULD NOT send payer identity types omitted in `payerIDs` record, none at all if record is not present.
 
-## Payer identity metadata record
+## 3. Committing payer to the invoice
 
-If `LN SERVICE` wants to get one or more types of payer identities from `LN WALLET` then it MUST put the following record into metadata which contains `application/payer-ids` string as first item further followed by records each indicating a type of identitiy that `LN WALLET` is allowed to provide. 
+If `SERVICE` requests (section 1) and `WALLET` sends a `payerid` record in the callback (section 2), the payer IDs must be committed to the metadata before the invoice is created.
```
hsjoberg:
```diff
--- If `SERVICE` requests (section 1) and `WALLET` sends a `payerid` record in the callback (section 2), the payer IDs must be committed to the metadata before the invoice is created.
+++ If `SERVICE` requests (section 1) and `WALLET` sends a `payerid` record in the callback (section 2), the payer IDs MUST be committed to the metadata before the invoice is created.
`

File: 18.md [hunk 8]
```diff
@@ -1,56 +1,96 @@
...
+```diff
+ [["text/plain", "description"], ["image/png;base64", "AAA=="]]
++[["text/plain", "description"], ["image/png;base64", "AAA=="]]{<identity records>}
 ```
 
-`LN WALLET` SHOULD NOT send payer identity types omitted in `application/payer-ids` metadata record, none at all if record is not present.
+After doing that, `SERVICE` can proceed to hash the metadata and include that hash in the generated BOLT11 invoice as the `descriptionHash` field.
 
-Upon seeing an `application/lnurl-auth` record with `isMandatory` set to `false` user SHOULD be explicitly given an option to include an `LNURL-AUTH` record in callback.
+On its own side, `WALLET` must do the same thing before hashing the metadata and checking its hashed value against the `descriptionHash` field of the invoice received from `SERVICE`.
```
hsjoberg:
```diff
--- On its own side, `WALLET` must do the same thing before hashing the metadata and checking its hashed value against the `descriptionHash` field of the invoice received from `SERVICE`.
+++ On its own side, `WALLET` MUST do the same thing before hashing the metadata and checking its hashed value against the `descriptionHash` field of the invoice received from `SERVICE`.
```

18.md
```diff
@@ -1,56 +1,96 @@
...
 
-Upon seeing an `application/lnurl-auth` record with `isMandatory` set to `true` user MUST be informed by `LN WALLET` that callback will contain an `LNURL-AUTH` record should a payment be sent.
+If `WALLET` sends more `payerid` fields than requested by `SERVICE`, `SERVICE` should still append everything as received.
```
hsjoberg:
```diff
--- If `WALLET` sends more `payerid` fields than requested by `SERVICE`, `SERVICE` should still append everything as received.
+++  If `WALLET` sends more `payerid` fields than requested by `SERVICE`, `SERVICE` SHOULD still append everything as received.

File: 18.md [hunk 10]
```diff
@@ -1,56 +1,96 @@
...
 
-When including an `LNURL-AUTH` record in a callback a `linkingKey` MUST be the one originally seen in `LNURL-PAY` static link.
+`WALLET` must not expect `SERVICE` to append the identity records to the metadata if `SERVICE` hadn't included the `payerIDs` record in the first response, as that would mean `SERVICE` doesn't understand LUD-18 at all.
```
hsjoberg:
```diff
--- `WALLET` must not expect `SERVICE` to append the identity records to the metadata if `SERVICE` hadn't included the `payerIDs` record in the first response, as that would mean `SERVICE` doesn't understand LUD-18 at all.
+++ `WALLET` MUST NOT expect `SERVICE` to append the identity records to the metadata if `SERVICE` hadn't included the `payerIDs` record in the first response, as that would mean `SERVICE` doesn't understand LUD-18 at all.
```

hsjoberg (changes requested):

fiatjaf:
> Why is there a mandatory flag everywhere again? This excludes a class of wallets who can't or don't want to deal with emails/ids. Mandatory is only justified for auth for its specific reasons.

It was @dpad85's solution and makes total sense. If a service wants to exclude itself from compatibility with non-supporting wallets that's their decision. In the same way a service might decide to exclude itself from compatibility with wallets that don't support lnurl-auth.

I imagine few situations in which these things will be really used, including the lnurl-auth bit (specially since there are so few wallets supporting that), but since we already have the `mandatory` stuff then better make it generic.

fiatjaf:
> I dislike how much fluff this adds. I don't see why we MUST force a set of those fields to be included ALL the time. The query string to send from WALLET becomes huge ALL the time..

Nothing is forced, that's the entire idea. Wallets can just ignore all the fluff, but I think it's really not that much.

Even for wallets that choose to support this, the querystring will likely contain just one payerid each time. Moreover this is not shown to user ever, so there is no issue, right?

Also, comparing this with the previous version of LUD-18 I think you'll agree this is much simpler to grasp.

18.md
dpad85: If `SERVICE` requires the payer's `name` and would also like their `email` (optional), the `payerIDs` field would be:

```
payerIDs: {
  "name": { mandatory: true },  <--- if not included by wallet, service will fail the payment
  "email": { mandatory: false } <--- just a nice to have
}
```

Other fields like `auth`  that `SERVICE` does not care about are simply not included.

dpad85:
> Why is there a mandatory flag everywhere again?

Basically, `payerIDs` has to be an array of `object` not of `string`, because `auth` is a complex object. So to be consistent, you'd write:
```
payerIDs: [ "email": ???, "identifier": ??? ]
```
Sure `???` could be some random filler data, but why not make it useful? The `mandatory` field does that somewhat. I can imagine a scenario where `SERVICE` makes `email` mandatory, so it has a way to contact the payer.

fiatjaf:
Going to merge this and we can improve it later.

Not ideal, but it's better than to keep the current LUD-18 there since it's already established that we're changing it.

PR #102, opened 2021-10-01
Visually separate wallets list

michaelWuensch:
This PR separates the wallets with "," to make it easier readable.

Also it removed the Alby duplicate in lud-4, the ThunderHub duplicate in lud-12 and Zap Android android from lud-11 (it does not support it)


fiatjaf:
Thank you very much.

PR #104, opened 2021-10-01
Remove ambiquity in lud-20

michaelWuensch:
This PR just changes one word. "MUST" instead of "SHOULD".
In my opinion SHOULD directly conflicts with lud-6.

In lud-6 WALLET is REQUIRED to show "text/plain".
lud-20 now says if the long description is not provided you should include "text/plain". But it would also be allowed to not show it.
Basically this means that a SERVICE that has not implement lud-20 but has implemented lud-6 cannot be sure his description makes it to a wallet that has implemented lud-20


hsjoberg:
Thank you for the fix!

PR #106, opened 2021-10-01
Lud-1: clarify case sensitivity

michaelWuensch:
In lud-1 the encoded LNURL is always displayed in UPPERCASE, but it is nowhere mentioned if this is required. This can lead to some people assuming UPPERCASE is mandatory.
I understand that for QR codes this makes sense, but for non QR codes the normal bech32 encoding is lowercase.

If someone checks for the fallback scheme and tests against "lightning=LNURL1" and a SERVICE created the fallback by using "lightning=lnurl1" it might actually not be recognized if it wasn't programmed in a case insensitive way.

I would suggest to allow both, uppercase and lowercase.
What do you think?


hsjoberg:
I mean this kinda relates to bech32 and the bech32 BIP which clarifies when uppercase and lowercase should be used.
But I suppose clarity doesn't hurt.

PR #122, opened 2021-10-08
LUD22: Webhook notification for `payRequest`

andrerfneves:
LUD22 allows for wallets to request that services also post a payload to a separate/external server. This allows for the creation of stateless LNURL interfaces that interact with a service (in this case through payRequest LUD6) without actually paying the invoice. The service will simply publish the information to the provided webhook URL.


22.md
```diff
@@ -0,0 +1,95 @@
...
+
+Webhook URLs must be SSL-enabled `https` URLs, and the `SERVICE` is encouraged to enable any and all restrictions and safety checks on their side. Additionally, [GET URL's accept around ~2000 characters for the entire request string](https://stackoverflow.com/a/417184). Therefore `webhook` query parameter can only be as large as to fit in the URL alongside any/all of the other properties outlined above.
+
+### Service-side
+
+When the invoice is settled, `SERVICE` must make a `POST` call to the `webhook` URL provided by the `WALLET` containing a body payload in the following format:
+
+```diff
+ {
+   invoice: String,
```
fiatjaf:
```suggestion
   paymentHash: String,
```

fiatjaf:
```suggestion
   paymentHash: String,
   preimage: String,
```

fiatjaf:
This is interesting, I just remembered I wanted to turn lntxbot into a stateless interface to lapps instead of being a full fledged wallet, and it would be tricky to do lnurlpay successAction things without something like this.

22.md
fiatjaf: So I suggested one thing and then another, but I actually think the webhook should contain _only_ the preimage, as that minimizes errors and confusion.

The hash can be computed from the preimage and that should be enough to prove an invoice was paid and also to match against an invoice db on the stateless wallet side.

fiatjaf:
It is not super stateless to have to have a server listening for webhooks (although it's easy to proxy webhooks into SSE events using things like pipedream.com), but asking services to offer a payment checking API endpoint (like @bumi wanted) is way too much and no one will implement -- while dispatching a single webhook is one line of code.

22.md
andrerfneves: Done

22.md
```diff
@@ -0,0 +1,58 @@
...
+### Note on Webhook URL parameters
+
+Webhook URLs must be SSL-enabled `https` URLs, and the `SERVICE` is encouraged to enable any and all restrictions and safety checks on their side. Additionally, [GET URL's accept around ~2000 characters for the entire request string](https://stackoverflow.com/a/417184). Therefore `webhook` query parameter can only be as large as to fit in the URL alongside any/all of the other properties outlined above.
+
+### Service-side
+
+When the invoice is settled, `SERVICE` must make a `POST` call to the `webhook` URL provided by the `WALLET` containing a body payload in the following format:
+
+```diff
+ {
```
hsjoberg, line 49: A bit pedantic perhaps, but this should probably be actual JSON

```diff
-{
-   paymentHash: String,
-   status: 'settled',
-   amount: MilliSatoshi,
-   preimage: String
- }
+{
+   "paymentHash": String,
+   "status": 'settled',
+   "amount": MilliSatoshi,
+   "preimage": String,
+}
```

22.md
```diff
@@ -0,0 +1,58 @@
...
+Upon seeing a `webhookAllowed` property set to `true` on the response from the `SERVICE`, a `WALLET` may then pass along a new query parameter named `webhook` to the second callback to `SERVICE`:
+
+```diff
+- <callback><?|&>amount=<milliSatoshi>
++ <callback><?|&>amount=<milliSatoshi>&webhook=<String>
+```
+
+### Note on Webhook URL parameters
+
+Webhook URLs must be SSL-enabled `https` URLs, and the `SERVICE` is encouraged to enable any and all restrictions and safety checks on their side. Additionally, [GET URL's accept around ~2000 characters for the entire request string](https://stackoverflow.com/a/417184). Therefore `webhook` query parameter can only be as large as to fit in the URL alongside any/all of the other properties outlined above.
```
hsjoberg, line 42: nit: Perhaps allow http on Onion service.

The GET URL limit kinda applies to every parameter and could be a notice in LUD-06 instead.

hsjoberg:
It looks like the PR is out of tree with the `lud` branch (probably because of rebase in the `lud` branch).
Best way to fix is most likely to just start your branch over from scratch.

andrerfneves:
Someone should probably fix this PR

andrerfneves:
Oh wait

andrerfneves:
Closing this

PR #139, opened 2022-01-05
Adding Galoy

leesalminen:
This PR adds Galoy to 01 and 06


fiatjaf:
Thank you, I'm missing a bunch of new updates here!'

PR #145, opened 2022-04-02
LUD-03: Add mermaid.js diagram to withdrawRequest

grunklejp:
I saw someone asking for this on twitter and thought I would write it up really quick. I can do this for the rest of the spec if there is interest in such things. 

> Note: I took the liberty of splitting up the Wallet & User (Alice) into separate entities because I've seen many cases of wallets not handling the user flow very well. (ie. Muun not allowing users to specify a withdrawal amount 😠 )

### DIFF
--- a/03.md
+++ b/03.md
@@ -2,6 +2,7 @@ LUD-03: `withdrawRequest` base spec.
 ====================================
 
 `author: akumaigorodski`
+`diagram: johnpaulkiser`
 
 ---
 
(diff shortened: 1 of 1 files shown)


hsjoberg:
Hi @johnpaulkiser. Interesting, thank you for this pull request.
I suggest we display the diagram in the end of the document so that the reader sees the spec text first.

> Muun not allowing users to specify a withdrawal amount 😠

Oof...

dni:
bump :D

PR #159, opened 2022-05-14
Remove LUD-12 support for Wallet of Satoshi

reneaaron:
Wallet of Satoshi does currently not support LUD-12. However, they are planning to implement it in Q2 or Q3.


hsjoberg:
LGTM.

PR #160, opened 2022-05-16
Remove the 512x512 size in the image specification

brianoflondon:
Remove the 512x512 size in the image specification as it read like a requirement.

Specification is a 100kb byte limit, 512x512 will most of the time be larger than this.


augustresende (approved):

augustresende (approved):

fiatjaf:
I think 100K is too much, we should reduce it to 40K.

PR #161, opened 2022-05-16
Add anchors to documents.

ibz:
This is so that one can, for example, link to https://github.com/fiatjaf/lnurl-rfc#user-content-lnurl-document-04 from a Lightning log in page and have the user immediately see the supported wallets.

Thank you!


README.md
```diff
@@ -5,26 +5,26 @@ These are all the individual documents describing each small piece of protocol t
...
+| [03](03.md) <a id="lnurl-document-03"></a> | `withdrawRequest` base spec.                                | [Balance of Satoshis][balanceofsatoshis], [Blixt][blixt], [BlueWallet][bluewallet], [Breez][breez], [coinos][coinos], [Fountain Podcasts][fountain], [LNbits][lnbits], [@lntxbot][lntxbot], [Muun][muun], [Phoenix][phoenix], [ShockWallet][shockwallet], [SimpleBitcoinWallet][sbw], [ThunderHub][thunderhub], [Wallet of Satoshi][wos], [Zap Android][zap], [Zap Desktop][zap], [Zap iOS][zap], [ZBD Discord][zbd], [ZBD Extension][zbd], [ZBD Telegram][zbd], [ZEBEDEE][zbd], [Zeus][zeus] |
+| [04](04.md) <a id="lnurl-document-04"></a>| Auth base spec.                                             | [Alby][alby], [Balance of Satoshis][balanceofsatoshis], [Blixt][blixt], [Breez][breez], [coinos][coinos], [LNbits][lnbits], [@lntxbot][lntxbot], [Phoenix][phoenix], [SeedAuth](seedauth), [SeedAuthExtension](sae), [SimpleBitcoinWallet][sbw], [ThunderHub][thunderhub], [Zap Desktop][zap], [Zeus][zeus] |
```
hsjoberg, line 11: Missing a space here

hsjoberg (changes requested):

hsjoberg:
Nice idea. Left a small nit.

README.md
```diff
@@ -5,26 +5,26 @@ These are all the individual documents describing each small piece of protocol t
...
-| [17](17.md) | Scheme prefixes and raw (non bech32-encoded) URLs.          | [Alby][alby], [Blixt][blixt], [BTCPayServer][btcp], [@lntxbot][lntxbot], [ZBD Discord][zbd], [ZBD Telegram][zbd] |
-| [18](18.md) | Payer identity in `payRequest` protocol.                    | [Blixt][blixt], [@lntxbot][lntxbot], [ZBD Discord][zbd], [ZBD Telegram][zbd] |
-| [19](19.md) | Pay link discoverable from withdraw link.                   | [Blixt][blixt], [SimpleBitcoinWallet][sbw] |
-| [20](20.md) | Long payment description for pay protocol.                  | [Alby][alby], [Blixt][blixt], [@lntxbot][lntxbot], [Phoenix][phoenix] |
+| [01](01.md) <a id="lnurl-document-01"></a> | Base LNURL encoding and decoding.                           | _all the ones listed below_ |
```
hsjoberg, line 8: Hmm how would it look with hashtag as the text for the tag? Otherwise it would be a hidden feature that few know about.

ibz: Hashtag?

hsjoberg:
```suggestion
| [01](01.md) <a id="lnurl-document-01">#</a> | Base LNURL encoding and decoding.                           | _all the ones listed below_ |
```

ibz: Note that this is not a link, but an anchor. It is `a` with `id`, not with `href`. It is not supposed to have any text because it is not linking anywhere, but rather it is used to be linked to.

hsjoberg: I see, yes it could be done to a link by just adding href too `<a href="#lnurl-document-01" id="lnurl-document-01">#</a>`...

I don't see the usefulness if no one would know about this feature.

ibz: You are suggesting to actually create a link to itself with "#" as a text? That would work, but it seems very strange. Have you ever seen that anywhere on the web?

What do you mean by "nobody would know about it"? This is how links work in HTML. Say, I want to link to a specific section inside a Wikipedia article, I would do it by using # and the ID of the element, like this: https://en.wikipedia.org/wiki/Padua#By_rail There is absolutely no visual indication that that element can be linked to, and yet people do this all the time - perhaps by using "view source" or the inspector? Or just guessing? I don't know. But there is no actual *link* in there, just a HTML element with an ID that can be linked to.

Anyway, I am happy to do whatever works for you, as long as we simplify the user experience. Currently, a web site doing Lightning log in, like stacker.news, will link to #lnurl-documents and it is completely not obvious for anyone that is not a Lightning developer that they need to look at "04".

Alternatively, we could maybe add a more user-friendly paragraph and link to that?

The problem now is that more and more websites are using Lightning log in, so we need to make it easily accessible to people who are using a Lightning wallet to log in for the first time.

No need to merge this if you don't like it, but at least I started a discussion about this, and hopefully there are other suggestions as well. https://lightninglogin.live/ is also nice BTW, but it seems more targeted to developers. Maybe something like that but with the user in mind would do, a page explaining Lightning log in for the final user, that anyone can link to from their log in pages? Perhaps hosted on GitHub Pages, so it doesn't disappear if we forget to renew the domain?

andrerfneves:
Closing as its super stale. We can improve this if we want to open another PR.

PR #162, opened 2022-05-25
Ibex support

cjc21:
IBEX Mercado supporting lud 03, 06, 16


fiatjaf:
Shouldn't this be added at the "services" section? Even though it has some functions that would be of a wallet, it's not really an actual personal wallet, right? I think it's saner to put it there.

cjc21:
> Shouldn't this be added at the "services" section? Even though it has some functions that would be of a wallet, it's not really an actual personal wallet, right? I think it's saner to put it there.

That makes sense. Done

fiatjaf:
Awesome, thanks!

PR #166, opened 2022-05-27
update CoinCorner luds

za-kk:


hsjoberg:
@za-kk Is removal of LUD-02 support correct?

za-kk:
Yes, we never supported LUD-02 so not sure how it made it in there anyway 😂

hsjoberg:
Haha I see. 😂

PR #175, opened 2022-08-02
Add an optional `wallet` query param to indicate to the `LN_SERVICE` which wallet was used to authenticate

steliosrammos:
Add an optional `wallet` query param to indicate to the `LN_SERVICE` which wallet was used to authenticate.


fiatjaf:
Is this a terrible thing that destroys privacy?

steliosrammos:
All that is shared is the wallet name. What are the privacy concerns you are thinking of?

Since it's an optional param, a wallet could also prompt a user before hitting the `LN_SERVICE` to ask if they want to share that info.

fiatjaf:
I don't know. I generally don't care about that kind of privacy, just raising the question.

benthecarman:
This could help fingerprint users, however, could be beneficial if we get into a world where some wallets support feature A,B,C and other wallets support X,Y,Z. Seems like PTLCs, BOLT12, AMP, Taro, etc are all moving this way

PR #184, opened 2022-10-04
Update dart_lnurl reference and new LUDs.

saentari:
Updated the dart_lnurl reference to the most up-to-date library.


fiatjaf:
This is awesome, so many LUDs!

PR #186, opened 2022-11-02
update CoinCorner luds

za-kk:
Updated CoinCorner luds to include LUD-19, also added CoinCorner to initial table of luds


fiatjaf:
Amazing.

This is for the CoinCorner wallet app, right?

za-kk:
Yes, we are experimenting with lud-19 on our bolt cards that can be controlled by the wallet app.

PR #188, opened 2022-11-10
Add Elixir lib

ramontayag:


fiatjaf:
This is an annoying request, but can you put the URL at the bottom and use the short link format like the others? Just so the entire table doesn't have to be realigned.

README.md
```diff
@@ -145,22 +145,23 @@ Self-hosted
...
+| [python-lnurl](https://github.com/python-ln/lnurl)                   | [01][01] [02][02] [03][03] [04][04] [06][06]                                                                                                             |
+| [bitcoin-s](https://github.com/bitcoin-s/bitcoin-s)                  | [01][01] [03][03] [06][06] [09][09]                                                                                                                      |
+| [rust-lnurl][rust-lnurl]                                             | [01][01] [03][03] [04][04]                                                                                                                               |
+| [spring-lnurl][springlnurl]                                          | [01][01] [04][04]                                                                                                                                        |
+| [lnurl_client Elixir][https://github.com/ramontayag/lnurl_client-ex] | [01][01] [06][06]                                                                                                                                        |
```
hsjoberg: This link is broken

```suggestion
| [lnurl_client Elixir](https://github.com/ramontayag/lnurl_client-ex) | [01][01] [06][06]                                                                                                                                        |
```

hsjoberg (changes requested):

README.md
ramontayag: 🤦 Sorry, didn't check! Will fix along with the [request](https://github.com/lnurl/luds/pull/188#issuecomment-1310072586) of @fiatjaf

ramontayag: Done - tested it on my branch as well.

ramontayag:
> This is an annoying request, but can you put the URL at the bottom and use the short link format like the others? Just so the entire table doesn't have to be realigned.

Done! 🙏

fiatjaf:
Thank you very much!

README.md
hsjoberg: Ah yes at the bottom is much better 🙏

PR #190, opened 2022-12-30
Full stop not Comma

HamishMacEwan:
"Full domain name here means FQDN with last comma omitted"

"Full domain name here means FQDN with last full stop omitted"


hsjoberg:
Thank you, @HamishMacEwan

PR #203, opened 2023-03-01
LUD-XX: LNURL Over Nostr

Kukks:
## Summary

Rewritten spec for LNURL-over-Nostr transport, replacing the original NIP-04 draft.

### Key changes from the previous version
- **NIP-04 → NIP-17**: All private exchanges use Gift Wrap (NIP-59 + NIP-44 encryption). NIP-04 is deprecated in the Nostr ecosystem.
- **`naddr` as primary addressing mode**: Service publishes LNURL params as a Kind 31120 parameterized replaceable event. Wallet fetches directly from the relay — no DM round trip, works even when the service is offline, and the `naddr` stays stable across param updates.
- **`nprofile` as fallback**: For session-specific LNURL types (auth with unique `k1`, etc.) that require an interactive DM exchange.
- **Removed**: `npub` support (unreliable without relay hints), "To Discuss" section (NIP-17 resolves the ephemeral-vs-async tradeoff — Gift Wraps are stored by relays for async delivery, but use ephemeral outer keys for privacy).
- **Added**: Error handling guidance, timeout recommendations, security considerations (key isolation, ephemeral wallet keys, replay protection), LUD-17 scheme parsing rules for `nostr:` URIs.

### Reference implementation

https://github.com/Kukks/LNURL (LNURL .NET — `ILNURLCommunicator` abstraction with `NostrLNURLCommunicator` + `LNURLNostrHelper`)

### DIFF
--- a/XX.md
+++ b/XX.md
@@ -1,0 +1,82 @@
+LUD-XX: LNURL Over Nostr.
+==============================================
+
+`author: kukks`
+
+---
+
+## Using LNURL over Nostr
+
+Many opponents of LNURL always point to the DNS requirement of LNURL as its biggest weakness, as it makes wallets require a publicly accessible http endpoint, often resulting in centralzied custodial solutions offering dominanely offering LNURL support.
+
+This LUD presents an alternative, where neither `SERVICE` nor `WALLET` require even an http server, by utilizing [Nostr](https://github.com/nostr-protocol).
+
+By utilizing [NIP-04](https://github.com/nostr-protocol/nips/blob/master/04.md)), [NIP-19](https://github.com/nostr-protocol/nips/blob/master/19.md)), [NIP-21](https://github.com/nostr-protocol/nips/blob/master/21.md)), and [NIP-33](https://github.com/nostr-protocol/nips/blob/master/33.md)), we can broadcast lnurl parameters and have its callbacks trigger the respective specification.
+
+  ### Various NIP19 scenarios
+  
+  NIP19 offers various note options that can point to public keys (`nprofile`), specific events (`nevent`), or parameters of a specific event (`naddr`). All 3 can be used to encode sufficient data to form a communication line for LNURL LUD-01 to function. We will also prefix with the uri `nostr:` scheme from NIP-21 to stay uri compliant within LNURL.
+
+#### Nprofile example
+When using `nprofile` for LNURL, it is assumed that the pubkey is exclusive for this lnurl usage.
+
+* Generate a new key for nostr: `b7226fb958fc69906a20ee5029bb15500f92f47136a5a4fff653a1186cc4ef3b`
+* Derive its public key: `fce2b7c9aa3019e65e808f47fe8cf99ae465103737f294a0444e6c5b53dee0c7`
+* Create nprofile (specifying `wss://r.x.com` as a relay hint) `nprofile1qqs0ec4hex4rqx0xt6qg73l73nue4er9zqmn0u555pzyumzm200wp3cpp4mhxue69uhhytnc9e3k7mgj4um4e`
+* Create nip21 uri : `nostr:nprofile1qqs0ec4hex4rqx0xt6qg73l73nue4er9zqmn0u555pzyumzm200wp3cpp4mhxue69uhhytnc9e3k7mgj4um4e`
+* Encode it in bech32 as per LUD01 `lnurl1dehhxarj8fh8qun0ve5kcef3w9chxvr9vv6xsetcx3e8z7ps0p6rvut8xuekcdende6k2dr9wguh5utddcc82df4x4c857t4d4ax6v3sxpmhqvmrwpcrgmtg0p6k2d3ew45xs7t5de3njefndvmk6em2x36k6dr9ynmp4m`
+* or as a url as per LUD17 `lnurlp:nprofile1qqs0ec4hex4rqx0xt6qg73l73nue4er9zqmn0u555pzyumzm200wp3cpp4mhxue69uhhytnc9e3k7mgj4um4e`
+
+You can now advertise this LNURL.
+
+* Wallet scans/loads LNURL
+* Wallet connects to the relay specified in nprofile
+* Wallet creates a new key for nostr and send a NIP4 to pubkey specified in nprofile, with empty content or a randomly generated string (to discuss)
+* Wallet watches for a reply to event id created in previous step from nprofile pubkey
+* Service creates a NIP4 event replying to wallet nostr event, with json parameters. In this case 
+```
+{
+  "callback": "nostr:nprofile1qqs0ec4hex4rqx0xt6qg73l73nue4er9zqmn0u555pzyumzm200wp3cpp4mhxue69uhhytnc9e3k7mgj4um4e",
+  "metadata": "[[\"text/plain\",\"hello world\"]]",
+  "tag": "payRequest",
+  "minSendable": 1000,
+  "maxSendable": 1000000,
+  "commentAllowed": 200
+}
+```
+Please note that the `callback` property is optional in this variant. If specified that next step is addressed to it, else it is addressed to service nostr event itself.
+* Wallet detects the event, reads parameters, and generates the query string parameters in the same format as usual, meaning `amount=1010&comment=whatever`
+* Wallet creates a new nostr event, with the query string parameters as its content. and replies to either the wallet nostr event, or to the callback nostr note definition. This callback helps prevent metadata leakage of public lnurl usage.
+* Service detects event, and send the lnurl response json back in NIP4 as a reply.
+* Wallet handles it as per lnurl specs
+
+#### NEvent example
+* Service creates a nostr event, where its content is the lnurl json, with event id being `c0e9e10110659220b25f83961d297194e596d2188943153918663d4904c478c7`
+* Services encodes it to nevent, `nevent1qqsvp60pqygxty3qkf0c89sa99cefevk6gvgjsc48yvxv02fqnz833cpp4mhxue69uhhytnc9e3k7mgalulyu` 
+* Create nip21 uri : `nostr:nevent1qqsvp60pqygxty3qkf0c89sa99cefevk6gvgjsc48yvxv02fqnz833cpp4mhxue69uhhytnc9e3k7mgalulyu`
+* Encode to lnurl bech32 or LUD17
+* Wallet scans/loads LNURL
+* Wallet connects to the relay specified in nevent
+* Wallet creates a new key for nostr
+* Wallet scans for event with id from nevent
+* Wallet detects the event, reads parameters, and generates the query string parameters in the same format as usual, meaning `amount=1010&comment=whatever`
+* Wallet creates a new nostr event, with the query string parameters as its content. and replies to either the wallet nostr event, or to the callback nostr note definition. This callback helps prevent metadata leakage of public lnurl usage.
+* Service detects event, and send the lnurl response json back in NIP4 as a reply.
+* Wallet handles it as per lnurl specs
+
+#### NAddr example
+
+Same as nevent, except that scanning for event requires using kind, author, and `d` identifier. 
+
+
+### Npub can also be used using the identical flow of `nprofile`, though no relay can be provided to guide the wallet where to connect to and would be less likely to find the event.
+
+### Callback property
+
+As mentioned, the `callback` (or other similar properties found for example in lnurl-auth) property is optional, and if left out, the callback mechanism is to tag the event that gave the parameters. 
+If however it is included, it could be set to an `nevent`, `nprofile`, `naddr`, `npub`. If the note had a public key, it would tag it in the event, if it had an event, it would tag it. This allows the Service to decouple the lnurl callback events, which works well when the lnurl is of npub or nprofile and reduces possible analysis of usage metrics.
+
+
+### To Discuss
+* Replace NIP4 with an ephemeral event?
+* Should we reduce the options as many duplicate functionality? To what?
--- a/XX.md
+++ b/XX.md
@@ -66,7 +66,7 @@ Please note that the `callback` property is optional in this variant. If specifi
 
 #### NAddr example
 
-Same as nevent, except that scanning for event requires using kind, author, and `d` identifier. 
+Same as nevent, except that scanning for event requires using kind, author, and `d` identifier. The benefits of using naddr over nevent, is that it allows you to update the parameters without needing to issue a new lnurl
 
 
 ### Npub can also be used using the identical flow of `nprofile`, though no relay can be provided to guide the wallet where to connect to and would be less likely to find the event.
--- a/XX.md
+++ b/XX.md
@@ -39,7 +39,7 @@ You can now advertise this LNURL.
 ```json
 {
 "callback": "nostr:nprofile1qqs0ec4hex4rqx0xt6qg73l73nue4er9zqmn0u555pzyumzm200wp3cpp4mhxue69uhhytnc9e3k7mgj4um4e",
-"metadata": "[["text/plain","hello world"]]",
+"metadata": "[[\"text/plain\",\"hello world\"]]",
 "tag": "payRequest",
 "minSendable": 1000,
 "maxSendable": 1000000,
--- a/XX.md
+++ b/XX.md
@@ -1,82 +1,70 @@
-LUD-XX: LNURL Over Nostr.
-==============================================
+LUD-XX: LNURL Over Nostr
 
-`author: kukks`
+Author: kukks
 
 ---
 
-## Using LNURL over Nostr
+## Introduction
 
-Many opponents of LNURL always point to the DNS requirement of LNURL as its biggest weakness, as it makes wallets require a publicly accessible HTTP endpoint, often resulting in centralized custodial solutions offering dominantly offering LNURL support.
+LNURL has a requirement for DNS, which some see as a major weakness, as it can lead to centralized solutions. This LUD offers an alternative to LNURL by using [Nostr](https://github.com/nostr-protocol/nostr), which does not require an HTTP server.
 
-This LUD presents an alternative, where neither `SERVICE` nor `WALLET` requires even an HTTP server, using [Nostr](https://github.com/nostr-protocol).
+## LNURL over Nostr
 
-By utilizing [NIP-04](https://github.com/nostr-protocol/nips/blob/master/04.md)), [NIP-19](https://github.com/nostr-protocol/nips/blob/master/19.md)), [NIP-21](https://github.com/nostr-protocol/nips/blob/master/21.md)), and [NIP-33](https://github.com/nostr-protocol/nips/blob/master/33.md)), we can broadcast lnurl parameters and have its callbacks trigger the respective specification.
+By using Nostr [NIP-04](https://github.com/nostr-protocol/nips/blob/master/04.md), [NIP-19](https://github.com/nostr-protocol/nips/blob/master/19.md), [NIP-21](https://github.com/nostr-protocol/nips/blob/master/21.md), it is possible to broadcast LNURL parameters and trigger the relevant specifications. This can be done without requiring an HTTP server for either the service or wallet.
 
-  ### Various NIP19 scenarios
-  
-  NIP19 offers various note options that can point to public keys (`nprofile`), specific events (`nevent`), or parameters of a particular event (`naddr`). All 3 can be used to encode sufficient data to form a communication line for LNURL LUD-01 to function. We will also prefix with the uri `nostr:` scheme from NIP-21 to stay uri compliant within LNURL.
+### Using NIP-19 for LNURL
 
-#### Nprofile example
-When using `nprofile` for LNURL, it is assumed that the pubkey is exclusive for this lnurl usage.
+NIP-19 has several options, but for this proposal, we will focus on "nprofile," which includes a public key and relay hints. This is sufficient to establish a communication line for LNURL [LUD-01](01.md)  to function.
 
-* Generate a new key for nostr: `b7226fb958fc69906a20ee5029bb15500f92f47136a5a4fff653a1186cc4ef3b`
-* Derive its public key: `fce2b7c9aa3019e65e808f47fe8cf99ae465103737f294a0444e6c5b53dee0c7`
-* Create nprofile (specifying `wss://r.x.com` as a relay hint) `nprofile1qqs0ec4hex4rqx0xt6qg73l73nue4er9zqmn0u555pzyumzm200wp3cpp4mhxue69uhhytnc9e3k7mgj4um4e`
-* Create nip21 uri : `nostr:nprofile1qqs0ec4hex4rqx0xt6qg73l73nue4er9zqmn0u555pzyumzm200wp3cpp4mhxue69uhhytnc9e3k7mgj4um4e`
-* Encode it in bech32 as per LUD01 `lnurl1dehhxarj8fh8qun0ve5kcef3w9chxvr9vv6xsetcx3e8z7ps0p6rvut8xuekcdende6k2dr9wguh5utddcc82df4x4c857t4d4ax6v3sxpmhqvmrwpcrgmtg0p6k2d3ew45xs7t5de3njefndvmk6em2x36k6dr9ynmp4m`
-* or as a url as per LUD17 `lnurlp:nprofile1qqs0ec4hex4rqx0xt6qg73l73nue4er9zqmn0u555pzyumzm200wp3cpp4mhxue69uhhytnc9e3k7mgj4um4e`
+#### Example of Nprofile
+
+When using nprofile for LNURL, it is assumed that the public key is exclusive to this LNURL usage. Here are the steps to create a nprofile:
+
+1. Generate a new key for Nostr: `b7226fb958fc69906a20ee5029bb15500f92f47136a5a4fff653a1186cc4ef3b`.
+2. Derive its public key: `fce2b7c9aa3019e65e808f47fe8cf99ae465103737f294a0444e6c5b53dee0c7`.
+3. Create nprofile specifying `wss://r.x.com` as a relay hint: `nprofile1qqs0ec4hex4rqx0xt6qg73l73nue4er9zqmn0u555pzyumzm200wp3cpp4mhxue69uhhytnc9e3k7mgj4um4e`.
+4. Create a NIP-21 URI: `nostr:nprofile1qqs0ec4hex4rqx0xt6qg73l73nue4er9zqmn0u555pzyumzm200wp3cpp4mhxue69uhhytnc9e3k7mgj4um4e`.
+5. Encode it in bech32 as per [LUD-01](01.md) `lnurl1dehhxarj8fh8qun0ve5kcef3w9chxvr9vv6xsetcx3e8z7ps0p6rvut8xuekcdende6k2dr9wguh5utddcc82df4x4c857t4d4ax6v3sxpmhqvmrwpcrgmtg0p6k2d3ew45xs7t5de3njefndvmk6em2x36k6dr9ynmp4m`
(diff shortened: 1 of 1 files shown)


cj-chua:
What a cool idea 🤯

Some thoughts:
Would it be worthwhile to make this LUD more restrictive by completely dropping lnurl-callback and using only `nprofile` in nip-19? Seems to me that it would make this LUD a lot simpler. The `SERVER` would simply dedicate a pubkey for this functionality and subscribe to kind4 on that pubkey

XX.md
```diff
@@ -0,0 +1,82 @@
...
+* Encode it in bech32 as per LUD01 `lnurl1dehhxarj8fh8qun0ve5kcef3w9chxvr9vv6xsetcx3e8z7ps0p6rvut8xuekcdende6k2dr9wguh5utddcc82df4x4c857t4d4ax6v3sxpmhqvmrwpcrgmtg0p6k2d3ew45xs7t5de3njefndvmk6em2x36k6dr9ynmp4m`
+* or as a url as per LUD17 `lnurlp:nprofile1qqs0ec4hex4rqx0xt6qg73l73nue4er9zqmn0u555pzyumzm200wp3cpp4mhxue69uhhytnc9e3k7mgj4um4e`
+
+You can now advertise this LNURL.
+
+* Wallet scans/loads LNURL
+* Wallet connects to the relay specified in nprofile
+* Wallet creates a new key for nostr and send a NIP4 to pubkey specified in nprofile, with empty content or a randomly generated string (to discuss)
+* Wallet watches for a reply to event id created in previous step from nprofile pubkey
+* Service creates a NIP4 event replying to wallet nostr event, with json parameters. In this case
```
Semisol: This should have a custom kind instead of kind 4

Semisol (changes requested):

arcbtc:
Also thought same for ages. Saves having to run server and doxing youself
https://t.me/nostr_protocol/15824

arcbtc:
This nostr atm uses nip4 as well. The DM it says to send to specific npub in the QR includes `encrypted-data, atm-Id`. The server gets the amount from `encrypted-data` (because its a fiat amount because the atm is offline), and sends a DM back asking for an invoice for x-amount. The user DMs the invoice and it is paid. 

This passing of additional data is usually done with url params, so a NostrLNURL would have to have the same ability!

XX.md
Kukks: I was thinking of reusing nip4 to mask usage of lnurl and make it appear as normal users messaging each other. That said, I was also debating if it should just be an ephemeral kind number and utilizing nip4 spec inside it

Kukks:
> What a cool idea 🤯
> 
> Some thoughts: Would it be worthwhile to make this LUD more restrictive by completely dropping lnurl-callback and using only `nprofile` in nip-19? Seems to me that it would make this LUD a lot simpler. The `SERVER` would simply dedicate a pubkey for this functionality and subscribe to kind4 on that pubkey

So the main drawback of profile is that it requires one more additional communication. The nevent/naddr provide an already available event with the lnurl protocol parameters. The callback is very useful as you would want to protect yourself from people analyzing all the calls to your pubkey. Competitors would know how many times your lnurl item is used very easily.

Maybe we can simplify it to only point to a npub (no need for profile as relay is already connected for initial handshake)

arcbtc:
Id say support both. Npub is easier and will be more widely used.

callebtc:
Been thinking about this a lot as well. What put me off was that you have two options:

- either you use kind 4 for the communication and "hide in the crowd" but also make social DM clients unhappy. We do this in Cashu but that doesn't require an immediate response which makes it easier to hide. An automated (i.e. fast) response to a kind 4 would be easy to detect.
- or you use another kind and then everyone can see who pays whom.

None of it is ideal IMO. Thoughts?

arcbtc:
Why do social clients matter? They shoud not assume DMs are for only their platform

arcbtc:
A wallet could use a different npub, perhaps thats the answer, people should not use keys for both payments and messaging

Kukks:
@callebtc  @arcbtc  I've greatly updated and simplified things by reducing many options that felt like feature creep.

The way this is designed, every lnurl should be a unique nostr key. As a user consuming a lnurl, I would advise generating an ephemeral key for each call, as you don't need that metadata clutter, and everything becomes linked to your main identity otherwise (very ethereum account ish which is bad)

Kukks:
Additionally, there is the callback section, where the lnurl can generate a unique key  for the callback so that you can for example, send the `amount` to without revealing you are asking for a bolt 11 invoice. The sender can even use a brand new key there too!

Kukks:
https://github.com/Kukks/LNURL/pull/4/

working branch on implementing this

Kukks:
> Been thinking about this a lot as well. What put me off was that you have two options:
> 
> - either you use kind 4 for the communication and "hide in the crowd" but also make social DM clients unhappy. We do this in Cashu but that doesn't require an immediate response which makes it easier to hide. An automated (i.e. fast) response to a kind 4 would be easy to detect.
> - or you use another kind and then everyone can see who pays whom.
> 
> None of it is ideal IMO. Thoughts?

Regarding
`We do this in Cashu but that doesn't require an immediate response which makes it easier to hide`

We can do this too! The other party does not need to be online, they can simply create a filter to scan for matching events when they are back!

XX.md
earonesty: nip4 has in-envelope tags.  just make a new tag..

Kukks: > nip4 has in-envelope tags. just make a new tag..

Not sure what you mean. Can you explain?

earonesty: nip4 defines 2 tags, "e" fir eventid and "p" for pubkey of receiver.

but yeah, a new kind is better.

relays should forward unknown kinds

Kukks: If a new kind is used, one can quickly scrape how many times a specific lnurl is used, exposing sensitive business information. Tags are also unencrypted in nip4, so that would also leak such data

fiatjaf:
Since this is an entirely different thing from LNURL I think this should be an opportunity for us to reevaluate things and fix the legacy crap LNURL has, maybe come up with something that is truly great. This, however, is just taking the crappy JSON from LNURL and putting it on Nostr.

Kukks:
> Since this is an entirely different thing from LNURL I think this should be an opportunity for us to reevaluate things and fix the legacy crap LNURL has, maybe come up with something that is truly great. This, however, is just taking the crappy JSON from LNURL and putting it on Nostr.

I'm open to it but what do you suggest to offer the least painful path to existing users and implementors? My current proposal is based on making this as minimal as possible (same structure and support for all lnurl sub specs) with the most amount of impact (no more HTTP endpoint requirement)

v0l:
This should use ephemeral-like events and should also require NIP-42 (AUTH) to prevent leakage of those events, encryption may not be required then.

As with HTTP you can't guarantee that the server is up or the Nostr-LNURL service is listening nor that the requester will still be there later so there is no benefit in storing these events.

Kukks:
> This should use ephemeral-like events and should also require NIP-42 (AUTH) to prevent leakage of those events, encryption may not be required then.
> 
> As with HTTP you can't guarantee that the server is up or the Nostr-LNURL service is listening nor that the requester will still be there later so there is no benefit in storing these events.

That relies on trust on relays which is not sufficient. mandatory ephemeral events would completely remcve the ability for processing lnurl asynchronously which could be a new exciting way to do new things. A sender/receiver can have state to resume processing of even requests in mid flow.

One can even use a master key for deterministic key generation for a flow so that one can store less data for such state

shafemtol:
I've been working on a new proposal for secure messaging over nostr with minimal public metadata. The main purpose I have in mind is secure DMs, but built in layers such that all kinds of application-specific messaging can use the same kind of messaging system and blend in with the rest. I think it would fit well with this use case.

First draft PR, which just defines how such events should look to the public (includes a kind for ephemeral messages): https://github.com/nostr-protocol/nips/pull/306

XX.md
```diff
@@ -0,0 +1,70 @@
...
+
+#### Example of Nprofile
+
+When using nprofile for LNURL, it is assumed that the public key is exclusive to this LNURL usage. Here are the steps to create a nprofile:
+
+1. Generate a new key for Nostr: `b7226fb958fc69906a20ee5029bb15500f92f47136a5a4fff653a1186cc4ef3b`.
+2. Derive its public key: `fce2b7c9aa3019e65e808f47fe8cf99ae465103737f294a0444e6c5b53dee0c7`.
+3. Create nprofile specifying `wss://r.x.com` as a relay hint: `nprofile1qqs0ec4hex4rqx0xt6qg73l73nue4er9zqmn0u555pzyumzm200wp3cpp4mhxue69uhhytnc9e3k7mgj4um4e`.
+4. Create a NIP-21 URI: `nostr:nprofile1qqs0ec4hex4rqx0xt6qg73l73nue4er9zqmn0u555pzyumzm200wp3cpp4mhxue69uhhytnc9e3k7mgj4um4e`.
+5. Encode it in bech32 as per [LUD-01](01.md) `lnurl1dehhxarj8fh8qun0ve5kcef3w9chxvr9vv6xsetcx3e8z7ps0p6rvut8xuekcdende6k2dr9wguh5utddcc82df4x4c857t4d4ax6v3sxpmhqvmrwpcrgmtg0p6k2d3ew45xs7t5de3njefndvmk6em2x36k6dr9ynmp4m`
```
hsjoberg: As we have a clean start, I suggest dropping LUD-01 and going only with LUD-17.

XX.md
```diff
@@ -0,0 +1,70 @@
...
+{
+"callback": "nostr:nprofile1qqs0ec4hex4rqx0xt6qg73l73nue4er9zqmn0u555pzyumzm200wp3cpp4mhxue69uhhytnc9e3k7mgj4um4e",
+"metadata": "[[\"text/plain\",\"hello world\"]]",
+"tag": "payRequest",
+"minSendable": 1000,
+"maxSendable": 1000000,
+"commentAllowed": 200
+}
+```
+Note that the `callback` property is optional in this variant. If specified, the next step is addressed to it; otherwise, it is addressed to the same Nostr LNURL.
```
hsjoberg: Seems like unnecessary asymmetry to make `callback` optional.
Mandating it to be the same nprofile again probably better and will make hard-typed languages happy.

hsjoberg (commented):
The title for this document is LNURL over Nostr.

But it seems like it's really LNURL-_pay_ over nostr.
Wouldn't it be better to support every LNURL subprotocol? They all see widespread usage.
And if not then I suggest making it clear that this is about LNURL-pay.

arcbtc:
I was assuming all lnurl was supported, which is how I can use for my atm. 

For the sanity of people implementing (although perhaps not optimal design), I cant help.thinking having it as similar as possible to LNURL the better, so they can just swap out the requests for nostr. ping pongs over ephemeral keys.

fiatjaf:
I don't think Nostr can work as a dumb network layer that relies on other people's computer. This proposal can only work if it done in a way that makes relays a central part of it, such that users can interact only with relays that have agreed somehow to do this -- for example, the receiver of a payment may be running their own relay or may have a relationship with some other relay that allows them to use the relay for sending and receiving these messages.

fiatjaf:
I think using Nostr has some clear extra features that HTTP doesn't have (while at the same time being worse than HTTP for some use cases), so if you just make HTTP-over-Nostr you'll just get the worst of both worlds, while if you do something that relies on Nostr properly you can get much better things.

Also this poses an opportunity for us to merge lnurl-withdraw and lnurl-pay into the same thing.

For example, the entire "withdraw from a service" flow could be done differently by having the service notify the user that their balance has updated instead of waiting for the user to scan a thing.

Both lnurl-withdraw and lnurl-pay and also "zaps" could be condensed into a single "payment offer" `kind:5000` event that is sent to the relays specified by the receiver. Then whenever the receiver is online and wants to actually receive that payment they can send an event back referencing the ids of the payment offers. The actual lightning invoice doesn't have to have a description hash since the Nostr event provides the necessary cryptographic commitments (although no one is interested in that).

For a service (or a person) that wants to sell stuff with lnurl or make lnurls that have stated purposes (i.e. they're not just donations or arbitrary payments, but they're actually tied to the delivery of some good or service) they can publish Nostr "sell offer" `kind:30010` events with some metadata that can be edited but also have a static "address" and they can share the address of these events directly with the potential payers (via QR-Code and whatnot). The payer then just makes a payment offer (as in the previous paragraph) referencing this.

The entire flow can happen asynchronously over a long period of time if both parties are ok with that and the relay keeps messages around.

Lightning Address in this scheme can be just a NIP-05 address. For example, I want to pay `hampus@blixtwallet.com`, I find his public key and relays from that address and use that make a payment offer, as in the paragraphs above. Of course this same flow works with `nprofile` too and even `npub` (even though that disgracefully doesn't have relays specified in it).

We can also have services selling stuff with Lightning Address (or nprofile/npub) by just having them publish their sell offers as events with `kind:30010`. The payer will query all these and decide which one to offer a payment for. This enables both someone to sell a single thing or even make a single tip jar with metadata -- but also more complex flows, for example, a pizza shop that wants to have a different sell offer for each of their pizza flavors. Then wallets can show all these as a menu inside their UIs.

XX.md
Kukks: It enhances privacy to be able to specify a completely different reply key.

XX.md
Kukks: I'm good with that 👍

Kukks:
> The title for this document is LNURL over Nostr.
> 
> But it seems like it's really LNURL-_pay_ over nostr. Wouldn't it be better to support every LNURL subprotocol? They all see widespread usage. And if not then I suggest making it clear that this is about LNURL-pay.

No, it supports all sub-protocols, I just included a singular example. It should work for all, and if I missed something please let me know so we make sure it supports everything. 

I'll make it more clear it's for all.

Kukks:
> I don't think Nostr can work as a dumb network layer that relies on other people's computer. This proposal can only work if it done in a way that makes relays a central part of it, such that users can interact only with relays that have agreed somehow to do this -- for example, the receiver of a payment may be running their own relay or may have a relationship with some other relay that allows them to use the relay for sending and receiving these messages.

Why? Connecting to a public relay and sending nip4 events back and forth is equivalent to all these twitterlike clients. In any case, the use of nprofile allows the lnurl creator to advertise their own relay if need be.

Kukks:
> I think using Nostr has some clear extra features that HTTP doesn't have (while at the same time being worse than HTTP for some use cases), so if you just make HTTP-over-Nostr you'll just get the worst of both worlds, while if you do something that relies on Nostr properly you can get much better things.
> 
> Also this poses an opportunity for us to merge lnurl-withdraw and lnurl-pay into the same thing.
> 
> For example, the entire "withdraw from a service" flow could be done differently by having the service notify the user that their balance has updated instead of waiting for the user to scan a thing.
> 
> Both lnurl-withdraw and lnurl-pay and also "zaps" could be condensed into a single "payment offer" `kind:5000` event that is sent to the relays specified by the receiver. Then whenever the receiver is online and wants to actually receive that payment they can send an event back referencing the ids of the payment offers. The actual lightning invoice doesn't have to have a description hash since the Nostr event provides the necessary cryptographic commitments (although no one is interested in that).
> 
> For a service (or a person) that wants to sell stuff with lnurl or make lnurls that have stated purposes (i.e. they're not just donations or arbitrary payments, but they're actually tied to the delivery of some good or service) they can publish Nostr "sell offer" `kind:30010` events with some metadata that can be edited but also have a static "address" and they can share the address of these events directly with the potential payers (via QR-Code and whatnot). The payer then just makes a payment offer (as in the previous paragraph) referencing this.
> 
> The entire flow can happen asynchronously over a long period of time if both parties are ok with that and the relay keeps messages around.
> 
> Lightning Address in this scheme can be just a NIP-05 address. For example, I want to pay `hampus@blixtwallet.com`, I find his public key and relays from that address and use that make a payment offer, as in the paragraphs above. Of course this same flow works with `nprofile` too and even `npub` (even though that disgracefully doesn't have relays specified in it).
> 
> We can also have services selling stuff with Lightning Address (or nprofile/npub) by just having them publish their sell offers as events with `kind:30010`. The payer will query all these and decide which one to offer a payment for. This enables both someone to sell a single thing or even make a single tip jar with metadata -- but also more complex flows, for example, a pizza shop that wants to have a different sell offer for each of their pizza flavors. Then wallets can show all these as a menu inside their UIs.



> 

While I like a lot of this (and mentioned something similar in the first original commit and [see here](https://iris.to/post/note1m7cg0e666f3qzfjjht9efdej6n578qah9g4em4cnwegsp5yp6n3s3u8dgw) ), I think we should first focus on feature parity for minimum disruption. I think we can plan what you are saying as a separate...NIPLUD (🤒 what have we done).

fiatjaf:
> Why? Connecting to a public relay and sending nip4 events back and forth is equivalent to all these twitterlike clients.

Yes, and they suck completely. They're completely non-scalable and they must change if Nostr is going to grow more. How on Earth are you going to know in which relay someone is publishing notes if there are a billion Nostr users? Or do you expect the top 10 relays to remain public and free and serving a billion users?

> While I like a lot of this (and mentioned something similar in the first original commit and see here ), I think we should first focus on feature parity for minimum disruption.

This is already an entirely new protocol with zero backwards-compatibility. Switching from the old LNURL to this will already require multiple man-hours on dozens of wallets and services. Calling Nostr relays is not as straightforward as calling HTTP endpoints. Everything is different, there are signatures, timeouts, event handling, different parsing. It makes no sense to think this will be a smooth change. We should at least make it a change worth doing by making a better protocol out of this.

> NIPLUD (￼ what have we done).

I guess it should be just a NIP. There is nothing about "URLs" on it anymore.

Kukks:
> How on Earth are you going to know in which relay someone is publishing notes if there are a billion Nostr users?
That's why we're using `nprofile` which provides relay hints.

I expect relays to start pruning old (that was stored for free) data. 

>This is already an entirely new protocol with zero backwards-compatibility. Switching from the old LNURL to this will already require multiple man-hours on dozens of wallets and services. Calling Nostr relays is not as straightforward as calling HTTP endpoints. Everything is different, there are signatures, timeouts, event handling, different parsing. It makes no sense to think this will be a smooth change. We should at least make it a change worth doing by making a better protocol out of this.

I disagree. This LUD adds an alternative communication layer and supports all current LUD interactions.  Is LNURL perfect? No. Can we come up with a better protocol? Probably. But we have this with very good adoption rates, and this PR allows additional adoption to the restricted areas in the usual, ad hoc way where nothing existing breaks.

I'm up for working on a better, brand new protocol in the future, but we need something now (as is evidenced by plenty of people trying to tape things together with no spec but in a very similar but uninteropable way)

Kukks:
> How on Earth are you going to know in which relay someone is publishing notes if there are a billion Nostr users?
That's why we're using `nprofile` which provides relay hints.

I expect relays to start pruning old (that was stored for free) data. 

>This is already an entirely new protocol with zero backwards-compatibility. Switching from the old LNURL to this will already require multiple man-hours on dozens of wallets and services. Calling Nostr relays is not as straightforward as calling HTTP endpoints. Everything is different, there are signatures, timeouts, event handling, different parsing. It makes no sense to think this will be a smooth change. We should at least make it a change worth doing by making a better protocol out of this.

I disagree. This LUD adds an alternative communication layer and supports all current LUD interactions.  Is LNURL perfect? No. Can we come up with a better protocol? Probably. But we have this with very good adoption rates, and this PR allows additional adoption to the restricted areas in the usual, ad hoc way where nothing existing breaks.

I'm up for working on a better, brand-new protocol in the future, but we need something now (as is evidenced by plenty of people trying to tape things together with no spec but in a very similar but un-interopable way)

fiatjaf:
> (as is evidenced by plenty of people trying to tape things together with no spec but in a very similar but uninteropable way)

Where is this happening? I am not aware know of any such efforts.

> Is LNURL perfect? No. Can we come up with a better protocol? Probably. But we have this with very good adoption rates, and this PR allows additional adoption to the restricted areas in the usual, ad hoc way where nothing existing breaks.

I don't see what you mean. Everything would break. A wallet supporting just HTTP-LNURL wouldn't work with a service supporting NOSTR-LNURL and vice-versa.

---

Anyway, I see you're trying to be pragmatic, but I don't feel any desire to work on any of this if it doesn't look like a good solution to me, I just feel depressed. If the bolt12 people would stop doing their bolt12 things and unite everybody in a single protocol I could be motivated to work on this -- or if we knew of some latent use case that would be open by the implementation of this imperfect solution that would expand Bitcoin's reach, anything like that.

But this I think will have no adoption, will confuse people, and get nothing done.

Kukks:
> > (as is evidenced by plenty of people trying to tape things together with no spec but in a very similar but uninteropable way)
> 
> Where is this happening? I am not aware know of any such efforts.
> 
> > Is LNURL perfect? No. Can we come up with a better protocol? Probably. But we have this with very good adoption rates, and this PR allows additional adoption to the restricted areas in the usual, ad hoc way where nothing existing breaks.
> 
> I don't see what you mean. Everything would break. A wallet supporting just HTTP-LNURL wouldn't work with a service supporting NOSTR-LNURL and vice-versa.
> 
> Anyway, I see you're trying to be pragmatic, but I don't feel any desire to work on any of this if it doesn't look like a good solution to me, I just feel depressed. If the bolt12 people would stop doing their bolt12 things and unite everybody in a single protocol I could be motivated to work on this -- or if we knew of some latent use case that would be open by the implementation of this imperfect solution that would expand Bitcoin's reach, anything like that.
> 
> But this I think will have no adoption, will confuse people, and get nothing done.

Ok, well, can you NACK this then we can move on?

fiatjaf:
I don't know how that NACK stuff works. I am just giving my opinion.

XX.md
Kukks: @hsjoberg Coming back to this, how would this work as LUD 17 specifies the lnurl protocol over the scheme, while I use the scheme to signal it is a nostr endpoint?

XX.md
earonesty: tags can be encrypted im the payload instead.   maybe i eas thinking of nip59

Kukks:
I am going to start pushing this once more.

Some points I'd like to discuss:
* Should there be a final, optional event sent by the receiver signaling a receipt? Could be public or private receipt, private by default by otherwise signalled public and would be a specific nostr kind with cleartext content.
* Should there be an initial, optional round signalling various payment protocols? Could be sending step3, but with a content stating "listpaymentmethods", in which case the receiver replies with an array/dictionary of options such as 
```
{
"lnurlp": "nostr:nprofile1..."
"bip21": "bitcoin:address?amount=1&pj="nostr:nprofile12"", // see https://github.com/Kukks/BTCPayServer.BIP78/tree/nostr/BTCPayServer.BIP78.Nostr
...
}
```

arcbtc:
GOOD!
I want this, and for the sanity of developers it should be lnurl, and GET/POST requests just nostr notes.
We could easily extend nip15 marketplaces to use, and I can finally build the "offline nostr powered bitcoin atm"

ok300:
>  if you just make HTTP-over-Nostr you'll just get the worst of both worlds, while if you do something that relies on Nostr properly you can get much better things.

> Calling Nostr relays is not as straightforward as calling HTTP endpoints. Everything is different, there are signatures, timeouts, event handling, different parsing. It makes no sense to think this will be a smooth change.

IMO fiatjaf is right.

Why not use NWC instead? Its nostr-native and covers LNURL-pay, LNURL-withdraw, even LNURL-check-balance.

shocknet-justin:
We're doing this more comprehensively at https://GitHub.com/shocknet/Lightning.Pub

We use a proto-gen to make it both rest and nostr compatible rpc.

An accounts solution is ultimately necessary as well to give it a full range of capabilities. We already power Lightning Video with it, and I welcome community involvement.

Kukks:
> > if you just make HTTP-over-Nostr you'll just get the worst of both worlds, while if you do something that relies on Nostr properly you can get much better things.
> 
> > Calling Nostr relays is not as straightforward as calling HTTP endpoints. Everything is different, there are signatures, timeouts, event handling, different parsing. It makes no sense to think this will be a smooth change.
> 
> IMO fiatjaf is right.
> 
> Why not use NWC instead? Its nostr-native and covers LNURL-pay, LNURL-withdraw, even LNURL-check-balance.

The issue I see is that every lnurl spec would need to be re-implemented here into the new protocol. Replacing the communication line is "trivial", at least in comparison to the adoption of a whole new protocol. I also would like to reuse the same basis and potentially allow other payments to use it but signaling other payment options, so that one can do Payjoin over nostr, which functions in the same way.





 

> We're doing this more comprehensively at https://GitHub.com/shocknet/Lightning.Pub
> 
> We use a proto-gen to make it both rest and nostr compatible rpc.
> 
> An accounts solution is ultimately necessary as well to give it a full range of capabilities. We already power Lightning Video with it, and I welcome community involvement.

I think this is beyond the scope of this feature, yours is basically a whole application stack for lightning

Sub7iam:
Hello
4 days ago i swaped doge to ltc after 45 minutes swap goes fail but my doges doesn't back to wallet , i try to contact and report trust wallet support team but they didn't answer me after 3 days.
Transaction hash goes to error 404 and i cant see transaction in blockchair.
would you mind telling me what iv to do please

Hash: 
https://blockchair.com/dogecoin/transaction/0f48bb5bcee306982ccc65cc720deaa3ae2eaf7ebdd048d5ccf66384f7f25b90?from=trustwallet

Regards

fiatjaf:
@Sub7iam unfortunately I can't help. I just think you should stop dealing with these shitcoins.

XX.md
oren-z0: This kind of URI misses the "route" information that lnurls contain. For example, two static lnurl-pay QR codes can refer to the same server: https://example.org/lnurlp/link1?hello=world , https://example.org/lnurlp/link2?size=big .
The current suggestion requires a new nsec/npub pair for every case.

I suggest not to use the NIP-21 URI. Instead, I suggest to encode a full url that could look like:
```
http://<nprofile...>.nostr1/<optional-route>?<optional-query-params>
```
This also goes along with @hsjoberg's suggestion to support LUD17 in the future, however I'm worried that LUD17 adoption could take a lot of time in general.
Clients already check the hostname's tld for ".onion" and use matching library for this case - so checking the tld for ".nostr1" and using another matching library will be pretty straight forward.

Notice that the tld is ".nostr1" and not just ".nostr". This will allow to support multiple protocols in the future. For example, [NIP-04](https://github.com/nostr-protocol/nips/blob/master/04.md) for direct messages is already not-recommended and will probably be replaced by [NIP-17](https://github.com/nostr-protocol/nips/blob/master/17.md).

For now, I think we should stick with the old direct-messages NIP-04, which is already implemented in many libraries.

XX.md
```diff
@@ -0,0 +1,70 @@
...
+You can now advertise this LNURL.
+
+### The Flow
+
+1. The wallet scans/loads the LNURL.
+2. The wallet connects to the relay specified in nprofile.
+3. The wallet creates a new key for Nostr and sends a NIP-4 to the public key specified in nprofile, with empty content or a randomly generated string.
+4. The wallet watches for a reply to the event ID created in the previous step from the nprofile public key.
+5. The service creates a NIP-4 event replying to the wallet Nostr event, with JSON parameters. For example:
+```json
```
oren-z0, line 39: All the lnurl protocols already work over http - so my suggestion is simply to wrap the http requests and responses as jsons - like implemented in [http2nostr](https://github.com/oren-z0/http2nostr) and [nostr2http](https://github.com/oren-z0/nostr2http): 
A request will look like:
```json
{
  "id": "<unique-request-id>",
  "url": "/route?queryparam=value",
  "method": "GET",
  "headers": { "user-agent": "...", "accept": "application/json" },
  "bodyBase64": "..."
}
```
Notice that I removed the "Host" header (we don't have a hostname).
Also, it would be simple to always encode the body in base64, for simple handling and debugging in case of unexpected binary/unicode data.

The response should look like:
```json
{
  "id": "<same-unique-id>",
  "status": 200,
  "headers": { "content-type": "application/json", ... },
  "bodyBase64": "..."
}
```

Kukks:
Fuck it. I'm bringing this back.

andrerfneves:
@Kukks you bringing this back or do we close?

PR #212, opened 2023-04-11
feat: add Alby wallet in lud-18 supported wallets

pavanjoshi914:
Alby supports lud-18 spec so here is the updates on the readme to showcase Alby in lud-18 

https://github.com/getAlby/lightning-browser-extension/blob/master/src/app/screens/LNURLPay/index.tsx#L88


pavanjoshi914:
cc @fiatjaf

fiatjaf:
You don't have to create a bunch of pull requests, you can force-push to the same branch of the original pull request.

PR #222, opened 2023-06-07
docs: add description for q query parameter

ChrisCho-H:
added guideline description for query parameter `q`, so that dev would not be confused.

### DIFF
--- a/01.md
+++ b/01.md
@@ -20,6 +20,7 @@ and presented as the following QR:
 ![A QR encoded LNURL example](https://i.imgur.com/HbB7U1K.png)
 
 Once `LNURL` is decoded:
+- `q` query parameter could be used as an identifier `k1` described in LUD02, LUD03 and so on
 - If `tag` query parameter is present, then this `LNURL` has a special meaning. Further actions will be based on `tag` parameter value.
 - Otherwise, a GET request should be executed, which must return a JSON object containing a `tag` field. Further actions will be based on `tag` field value.
 


fiatjaf:
This is not correct, and got me confused.

ChrisCho-H:
As I understand,
In a situation where A wallet tries to withdraw to B wallet using lnurl is like
1. A wallet gives encoded lnurl like `lnurl1...`
2. B wallet  decodes it(which would be `https...?q=abc...`) and request get info
3. A wallet use query param `q` to identify the request and return identified withdrawal request info like
```javascript
{
    "tag": "withdrawRequest", // type of LNURL
    "callback": string, // The URL which LN SERVICE would accept a withdrawal Lightning invoice as query parameter
    "k1": string, // Random or non-random string to identify the user's LN WALLET when using the callback URL
    "defaultDescription": string, // A default withdrawal invoice description
    "minWithdrawable": number, // Min amount (in millisatoshis) the user can withdraw from LN SERVICE, or 0
    "maxWithdrawable": number, // Max amount (in millisatoshis) the user can withdraw from LN SERVICE, or equal to minWithdrawable if the user has no choice over the amounts
}
```
and in this step, k1 value is actually same with q query parameter.

and also same with k1 used in callback in next step

```
<callback>
  <?|&> // either '?' or '&' depending on whether there is a query string already in the callback
   k1=<k1> // the k1 specified in the response above
  &pr=<lightning invoice> // the payment request generated by the wallet
```

@fiatjaf in what point is not correct? I'm just wondering.

fiatjaf:
There is no `q` parameter anywhere, this is not specified. You can use a `q` parameter if you want, but this is your choice, shouldn't be in the spec. You can use any parameters you desire.

ChrisCho-H:
alright. thx!

PR #227, opened 2023-07-16
Add BitBanana to LUD-20 supported wallets

michaelWuensch:
Since v0.6.5, which was just released, BitBanana supports lud-20.


hsjoberg:
Epic!

PR #234, opened 2023-08-25
LNURL-pay: Drop metadata description hash validation

callebtc:
This PR removes the `description_hash`requirement in LUD06 and LUD18.

## Background

LNURL-pay requires the `SERVER` to provide an invoice with a specific `description_hash` (`metadata` in LUD06, and `metadata+payerData` in LUD18). `WALLET` is then supposed to verify that the invoice it receives indeed has this `description_hash`. 

## Rationale

If I understand correctly, the original intention of this is to make the server commit to the receiver (or receiver+payer) identities. Practically, this doesn't really improve anything: If there was a MITM changing the invoice, they could do so since neither of the data committed to is secret. If the server itself is malicious, they could fake everything as they wish anyway.

Implementing this (possibly redundant) feature is also the biggest challenge when implementing LNURL-pay. 

This will also once and for all solve all the issues we have had for years dealing with CLN's requirement to provide the entire description at invoice creation.

I propose to drop this requirement, starting with removing the check `WALLET` performs on the invoice received in the second LNURL-p response. `SERVER`'s who don't want to upgrade, can remain as is.

### Supported

- Blixt [PR](https://github.com/hsjoberg/blixt-wallet/issues/1280) ✅
- Alby [PR](https://github.com/getAlby/lightning-browser-extension/pull/2700) ✅
- Zeus [PR](https://github.com/ZeusLN/zeus/pull/1676) ✅
- Current [Comment](https://github.com/lnurl/luds/pull/234#issuecomment-1694267678) ✅
- BlueWallet [PR](https://github.com/BlueWallet/BlueWallet/pull/6074) ✅
- Stacker.News [PR](https://github.com/stackernews/stacker.news/pull/775) ✅
- lnurl-go [PR](https://github.com/fiatjaf/go-lnurl/issues/19) ✅
- Phoenix [Commit](https://github.com/ACINQ/phoenix/commit/4334ef78de8b2ad53f02b79f3bb48f6162909032) ✅
- LNbits [PR](https://github.com/lnbits/lnbits/pull/1951) ✅
- Breez SDK [PR](https://github.com/breez/breez-sdk/pull/755) ✅
- Galoy Blink [PR](https://github.com/GaloyMoney/galoy/issues/3639) ✅
- SatSale [PR](https://github.com/SatSale/SatSale/issues/157) ✅
- sudonym-btc [PR](https://github.com/sudonym-btc/zap/issues/1) ✅
- LightningTipBot [Commit](https://github.com/LightningTipBot/LightningTipBot/commit/52e1fc043a4f0e348ecba0aa458f501d178fe5e4) ✅
- Geyser [PR](https://github.com/geyserfund/geyser-app/issues/1279) ✅
- Wallet of Satoshi (tested) ✅
- Mutiny [Comment](https://nostr.com/nevent1qqsyv62429vfmcnwy8wsxmn7mnzpj7hg5wh4rf4f8rn8q6gqcpkdurgpz4mhxue69uhkummnw3ezummcw3ezuer9wchsyg8plualm48yqv2etxcgkn7vsfz74fg5vdlp6nkz4ctxkapngxlp4upsgqqqqqqszxu2d7) ✅

### Unclear
_(please report PRs or test results for more wallets)_
- Green
- ...

### DIFF
--- a/06.md
+++ b/06.md
@@ -7,7 +7,7 @@ LUD-06: `payRequest` base spec.
 
 The idea here is that a wallet can scan a static QR code or click on a static LNURL address and get back details about a payment that is expected. The details may include much more extensive metadata than a normal Lightning invoice. And the amounts may be fixed or within a range.
 
-Then, once the user accepts the terms (and choose an amount, if that is not fixed), the wallet will call the service and get a Lightning invoice specific for that payment, containing a hash of the metadata as its [`h` tag (`description_hash`)](https://github.com/lightningnetwork/lightning-rfc/blob/master/11-payment-encoding.md#requirements-3) and proceed to pay the invoice if it matches the expected amount and hash.
+Then, once the user accepts the terms (and choose an amount, if that is not fixed), the wallet will call the service and get a Lightning invoice specific for that payment and proceed to pay the invoice if it matches the expected amount and hash.
 
 ## Pay to static QR/NFC/link
 
@@ -22,7 +22,7 @@ Then, once the user accepts the terms (and choose an amount, if that is not fixe
         "callback": string, // The URL from LN SERVICE which will accept the pay request parameters
         "maxSendable": number, // Max millisatoshi amount LN SERVICE is willing to receive
         "minSendable": number, // Min millisatoshi amount LN SERVICE is willing to receive, can not be less than 1 or more than `maxSendable`
-        "metadata": string, // Metadata json which must be presented as raw string here, this is required to pass signature verification at a later step
+        "metadata": string, // Metadata json which must be presented as raw string here
         "tag": "payRequest" // Type of LNURL
     }
     ```
@@ -69,7 +69,7 @@ Then, once the user accepts the terms (and choose an amount, if that is not fixe
     "[[\"text/plain\", \"lorem ipsum blah blah\"]]"
     ```
 
-3. `LN WALLET` displays a payment dialog where user can specify an exact sum to be sent which would be bounded by:
+4. `LN WALLET` displays a payment dialog where user can specify an exact sum to be sent which would be bounded by:
 
     ```
     max can send = min(maxSendable, local estimation of how much can be sent from wallet)
@@ -82,7 +82,7 @@ Then, once the user accepts the terms (and choose an amount, if that is not fixe
     And it may include:
     - An image element with the contents of `image/png` or `image/jpeg`.
 
-4. `LN WALLET` makes a GET request using
+5. `LN WALLET` makes a GET request using
 
     ```
     <callback><?|&>amount=<milliSatoshi>
@@ -90,7 +90,7 @@ Then, once the user accepts the terms (and choose an amount, if that is not fixe
 
     `amount` being the amount specified by the user in millisatoshis.
 
-5. `LN Service` takes the GET request and returns JSON response of form:
+6. `LN Service` takes the GET request and returns JSON response of form:
 
     ```Typescript
     {
@@ -105,17 +105,5 @@ Then, once the user accepts the terms (and choose an amount, if that is not fixe
     {"status":"ERROR", "reason":"error details..."}
     ```
 
-6. `LN WALLET` Verifies that `h` tag in provided invoice is a hash of `metadata` string converted to byte array in UTF-8 encoding.
 7. `LN WALLET` Verifies that amount in provided invoice equals the amount previously specified by user.
-10. `LN WALLET` pays the invoice, no additional user confirmation is required at this point.
-
-## Notes on metadata for server-side LNURL-PAY
-
-### When client makes a first call:
-
-Construct a metadata object, turn it into json, then include it into parent json as a string.
-
-### When client makes a second call
-
-1. Make a hash as follows: `sha256(utf8ByteArray(unescaped_metadata_string))`.
-2. Generate a payment request using an obtained hash.
+8. `LN WALLET` pays the invoice, no additional user confirmation is required at this point.
--- a/06.md
+++ b/06.md
@@ -7,7 +7,7 @@ LUD-06: `payRequest` base spec.
 
 The idea here is that a wallet can scan a static QR code or click on a static LNURL address and get back details about a payment that is expected. The details may include much more extensive metadata than a normal Lightning invoice. And the amounts may be fixed or within a range.
 
-Then, once the user accepts the terms (and choose an amount, if that is not fixed), the wallet will call the service and get a Lightning invoice specific for that payment and proceed to pay the invoice if it matches the expected amount and hash.
+Then, once the user accepts the terms (and choose an amount, if that is not fixed), the wallet will call the service and get a Lightning invoice specific for that payment and proceed to pay the invoice if it matches the expected amount.
 
 ## Pay to static QR/NFC/link
 
--- a/18.md
+++ b/18.md
@@ -7,8 +7,6 @@ LUD-18: Payer identity in `payRequest` protocol.
 
 The idea here is that the payer can identify itself when paying. The scope varies from free-form names (or [LUD-16](16.md) internet identifiers) useful for loose identification in comments or similar things, to cryptographic keys for proof of payment, to authentication keys for future login into a `SERVICE`'s website after a payment and other use cases.
 
-Aside from that, the payer ids are also committed to the invoice by means of the `descriptionHash` property, which ensures strong cryptographic guarantees for proof of payment.
-
 ## 1. `payerData` record
 
 If `SERVICE` wants to get one or more types of payer identities from `WALLET` then it MUST alter its JSON response to the first callback to include a `payerData` field, as follows (notice that the `payerData` record below has a bunch of fields only for completion, an actual response will likely contain just a subset of these):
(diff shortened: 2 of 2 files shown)


fiatjaf (approved):

Kukks:
ACK from my end. Less hassle dealing with LN implementation support and much easier to build some more complex flows.

hsjoberg (approved):

Egge21M:
ACK. Current Wallet, in foresight of this PR, never implemented this check.

michael1011 (approved):

kaloudis:
ACK. Will support this change in Zeus.

bumi:
ACK
this makes it easier for wallet developers and it makes it also easier for services to offer LNURL/lightning address with different node backends.

Alby will support this in the next release.

arcbtc:
Awesome, description_hash got in the way a bunch

callebtc:
Looks like there is broad support for this PR! Anything that should be considered before it can be merged?

dni:
:+1:

fiatjaf:
My only worry is that services will drop support for this too soon and start breaking wallets that haven't yet.

dpad85:
ACK, committing the lnurl-pay metadata to the invoice does not bring additional security AFAICT. The service is trusted anyway. Next Phoenix release will remove this check.

kristapsk (approved):
ACK, this would simplify things for SatSale and also allow implementing LN address for LndHub backend.

callebtc:
Can we merge this or are there any blocking issues @fiatjaf?

hsjoberg:
Is there are any value to commit [LUD-18](https://github.com/lnurl/luds/blob/luds/18.md) data to the invoice? This proposal removes desc hashing altogether.

NCrashed:
I would like to offer some critical feedback on the proposed change, particularly focusing on a crucial aspect of the [lud06](https://github.com/lnurl/luds/blob/luds/06.md): `9. LN WALLET pays the invoice, no additional user confirmation is required at this point`. The absence of a description hash eliminates the connection between the invoice and the user's explicit agreement, leaving the user without any evidence of discrepancies or potential misconduct by the server. A practical application of metadata commitment, from my experience, involves embedding exchange order numbers within this field, underscoring its significance. This issue becomes even more critical with `payerData` in [lud18](https://github.com/lnurl/luds/blob/luds/18.md), as this data originates from the client, not the server.

In summary:

* The principle of commitment is vital for ensuring users explicitly authorize transactions involving their funds.
* The removal of a cryptographic link between the request and the invoice opens up new vulnerabilities for server attacks, leaving users without verifiable evidence of their payments.
* Excluding the commitment mechanism compromises other components of the LUD, especially since the metadata format was specifically designed to facilitate hash computation and minimize verification ambiguities.

As an alternative, I suggest making the description hash optional and introducing an additional step for user confirmation when the commitment is absent.

evd0kim:
NACK. I do not see a single valid argument here.

Joining to @NCrashed.

I don't care about imaginary services and wallets that would benefit from doing less work. The developer's job is to ensure that the digest corresponds to the received data.

The current protocol ensures that the previous round of communication corresponds to the next one, and the wallet checks for the user in the background that THIS invoice is for THAT menu inputs + information.

The thesis about trusted/non-trusted service is entirely irrelevant here. If meta contained some important info there is no basis for building trust for random user-service as well as no hash to prove misbehaving or indicating for the user that the service showed one info and asked to pay for another. 

> MITM changing the invoice, they could do so since neither of the data committed to is secret. 

The check is not related to invoices but rather to any data put into Metadata.

> If the server itself is malicious, they could fake everything as they wish anyway.

Yes. If the service wants to receive payment and doesn't provide any services or goods, the protocol is irrelevant, but this change enables a whole spectrum of situations when the service may attempt to cheat just a little bit. For example, let's consider the original purpose of the LNURL pay protocol. Selling channel liquidity. The service may provide in Metadata specific parameters about channel fees, channel size and active period, and node pubkey. It receives the payment for the service and opens the channel of size not 1.5M but 1.499M, fees not 1000 ppm but 1001 ppm. Now good luck catching such dishonest service in every UI of every wallet you meant while saying "it will make everything much easier".  Instead, Metadata could be shown once, and `description_hash` could be checked in the background.

callebtc:
> The absence of a description hash eliminates the connection between the invoice and the user's explicit agreement

Could you be specific? What does this prevent in your view? If you could make an explicit example of how this improves a payment flow or how it could be abused, it might be easier to understand what you mean.

In my view, if I offer you an invoice, you can choose to pay it or not, that is the only necessary expression of your agreement. The proof of payment is the preimage you'll receive.

What else other than authenticated transport (https) do you need to be sure that this particular invoice is from the server you've actually asked the invoice from? And what misconduct does the payment hash prevent?

callebtc:
> The service may provide in Metadata specific parameters about channel fees, channel size and active period, and node pubkey.

This is a good example of a case that almost nobody cares about but it could be absolutely implemented even if the requirement would be dropped. If you run such a service, put the terms of the contract into the description hash and make your user read the contract and explicitly agree to it. That's the only way to make sure that both parties explicitly agree.

I'm guessing that 99.99% of lnurl payments are without any explicit commitments that the user cares about, they just want to pay. I haven't observed a single case where a dispute was resolved by comparing descriptions and hashes. For those cases where it's relevant: just do it.

evd0kim:
> Could you be specific? What does this prevent in your view? If you could make an explicit example of how this improves a payment flow or how it could be abused, it might be easier to understand what you mean.

I provided an example. 

> In my view, if I offer you an invoice, you can choose to pay it or not, that is the only necessary expression of your agreement. The proof of payment is the preimage you'll receive.

Checking a hash against a description that is stored or provided off-band may be more relevant for automated payments. It may be important for the client side, too. 

> What else other than authenticated transport (https) do you need to be sure that this particular invoice is from the server you've actually asked the invoice from? And what misconduct does the payment hash prevent?

The point being made is not about faking the server. The point being made is about the server that provides some imprecise/non-deterministic data.

> If you run such a service, put the terms of the contract into the description hash and make your user read the contract and explicitly agree to it. That's the only way to make sure that both parties explicitly agree.

Not relevant at all for machine-machine interfaces. 

> I haven't observed a single case where a dispute was resolved by comparing descriptions and hashes.

It wouldn't happen with the former rules, even if it were the case. There would not be any dispute at all.

johnzweng:
Hi! 🙂 

Is there any chance that this could be merged. Or if this is still a concern:

> My only worry is that services will drop support for this too soon and start breaking wallets that haven't yet.

Could we then add at least a deprecation note to the spec? I was not aware of this PR and (as a sending service) implemented this check and was wondering why more and more receivers (LN address owners) do not provide the correct tag `h` in their returned Bolt11 (effectively blocking the payment and creating support costs for our customer care).

Until someone (@reneaaron , thanks for that 🙏 ) pointed me to this PR here. 🙈

michaelWuensch:
Merging the PR would really make sense. Wallets that implement the spec as is already experience payment failures.
@callebtc  I just adopted this change in BitBanana (https://github.com/michaelWuensch/BitBanana/commit/cc59fec5d1c28f7418079b733267772d35486c78)

dni:
> Hi! 🙂
> 
> Is there any chance that this could be merged. Or if this is still a concern:
> 
> > My only worry is that services will drop support for this too soon and start breaking wallets that haven't yet.
> 
> Could we then add at least a deprecation note to the spec? I was not aware of this PR and (as a sending service) implemented this check and was wondering why more and more receivers (LN address owners) do not provide the correct tag `h` in their returned Bolt11 (effectively blocking the payment and creating support costs for our customer care).
> 
> Until someone (@reneaaron , thanks for that 🙏 ) pointed me to this PR here. 🙈

i agree with adding a deprecating note to the spec, but i wouldn't remove that feature completely, how about making the check optional and the users signal that he want to have it checked, default is no check?

evd0kim:
Valet now supports both protocols and it is able to check hash against description.

image

At least recently I've seen few more use cases for this feature, mostly for cross-protocol capabilities, so I keep telling it was a mistake to remove this from the spec.

evd0kim:
It is even wrong from the start to me.

> If I understand correctly, the original intention of this is to make the server commit to the receiver (or receiver+payer) identities. 

Hash to Description is a commitment to the intent/contract/information provided earlier and not to identities. So whole message somehow irrelevant.

michaelWuensch:
@evd0kim @dni 
To be honest, I do not understand what security benefit the descriptionHash verification has. There is no commitment to anything. The service can set the hash correctly in the invoice so that the verification succeeds and still do something completely different. So its no protection against fraud. Or do I miss something? 

Now to the suggestion to support both.
Maybe removing the verification was unnecessary. But shifting the burden to the user is an absolute UX nightmare in my opinion. Lightning wallets are already complicated enough, we don't need warnings that make users feel insecure and require a deep dive into some tech specs to understand.

Don't get me wrong, if this is a real security issue and I just don't get it, then I strongly support a warning. But if it is just about signaling an intention and only the intention can be cryptograpically verified, but not the final action, we are better off without a warning.

dni:
https://github.com/lnurl/luds/pull/96 so this is one usecase of lud-18

johnzweng:
Hi! 🙂 

I'm on board with @michaelWuensch on this. I don’t see how setting the `h` tag of the invoice to some hash provides any real security improvement.

> Hash to Description is a commitment to the intent/contract/information provided earlier and not to identities.

Yes, it acts as a commitment—but it’s not a _binding_ commitment. Since no secret is involved, _any_ lightning node could recreate the same commitment. If the **LN SERVICE** (LNURL web server) were compromised and returned a malicious invoice from a different node, it could still include the same `h` tag without issue. (And this is also true for the lud-18 flow.)

This suggests that preventing such an attack is _not_ the purpose of this feature. So, what is the actual attack scenario which gets prevented by adding this tag `h` to the invoice?

michaelWuensch:
The only benefit I can see is that in a lud-18 flow you have proof that the service actually received the identity data you sent. 
But if the wallet implementation is correct, this is always the case.

johnzweng:
I think it's the same when looking to @evd0kim's example from above:

> For example, let's consider the original purpose of the LNURL pay protocol: selling channel liquidity. The service may provide specific parameters in the metadata, such as channel fees, channel size, active period, and node pubkey. It receives the payment for the service but then opens the channel with a size of 1.499M instead of 1.5M and charges 1001 ppm instead of 1000 ppm.

I’d really love to understand how the `h` tag in the invoice could effectively prevent this scenario. The liquidity selling node could still return such invoice which commits to the correct specific agreement (channel parameters) but then do something totally different. So what have we gained in this situation by adding the `h` tag?

It just gives the payer the security that the receiver has actually _received_ the data correctly (without any guarantees what the receiver does with this information or adheres to it). But for "_just to be sure that data was transmitted correctly_" we have other layers (https, tcp). 

 Do I miss something? 🙂

evd0kim:
@johnzweng 

> Since no secret is involved

If the host or user puts a secret in description it may contain a secret. 

> This suggests that preventing such an attack is not the purpose of this feature. 

I am not talking about attacks per se. But generally yes, removing description hash check somehow makes protocol less secure if you mean some information provided via description off band. 

> something totally different. 

The point of channel opening example is that there is a payment amount that is fixed in BOLT11 against description with hash. It may prevent something only when wallet does checks. And when it does and something is wrong, the wallet user may prove it otherwise she couldn't. 

@dni 

Thanks for the link. Looks like LNURL withdraw scheme.

@michaelWuensch 

> Maybe removing the verification was unnecessary. But shifting the burden to the user is an absolute UX nightmare in my opinion. Lightning wallets are already complicated enough, we don't need warnings that make users feel insecure and require a deep dive into some tech specs to understand.

Well, this is the point of hashes. They should match on the background. I believe LN is capable to facilitate smart contracts logic on top, this PR moves it in entirely opposite direction making LNURL less capable and less "smart".

Kukks (approved):

andrerfneves (approved):
gg

PR #236, opened 2023-09-13
add bitlifi wallet to README.md

gorrdy:
Wallet Bitlifi now supports these luds, 01, 03, 06, 16


hsjoberg:
Great!

andrerfneves:

PR #242, opened 2023-10-05
LUD-XX Chained LNURLs

arcbtc:
For our Satsdice application, a user pays LNURL-pay and "might" receive a withdraw if they win. Currently they have to open a link to the LNURL-withdraw, it would be much better if the wallet automated this, so the wallet either went into the LNURL-pay workflow again (if they lost, so they can play again) or the LNURL-withdraw workflow if they won.

All LNURLs could be chained together through the response.

(I was inspired into action by BitcoinEkasi, they also have a good use case.)

### DIFF
--- a/24.md
+++ b/24.md
@@ -1,0 +1,8 @@
+WIP: LUD-24: LNURL chain.
+=====================================================
+
+`author: arcbtc`
+
+---
+
+Extends LUD-19, so ANY LNURL can be returned in the JSON response.
--- a/24.md
+++ b/24.md
@@ -5,4 +5,17 @@ WIP: LUD-24: LNURL chain.
 
 ---
 
-Extends LUD-19, so ANY LNURL can be returned in the JSON response.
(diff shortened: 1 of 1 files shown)


ghost:
Many time and attendance solutions primarily depend on strict rules and penalties, which can contribute to an unfavourable work environment and may not effectively motivate employees to meet attendance expectations. We advocate for a shift in focus from punitive systems to rewards-based systems, as we believe that positive reinforcement could lead to more favourable outcomes. We aim to encourage punctual attendance through SATS rewards by developing an LN Bits extension for time and attendance tracking.

dni:
i think i have a cleaner solution for the satsdice usecase. its almost like the `AesAction`  but called `WithdrawAction` you send a encrypted payload in the `LnurlPayActionResponse` which is the k1 of a withdraw, the wallet can decrypt the withdraw payload with the preimage (like the `AesAction` and then can immediatly withdraw via lud08 fastwithdraws. https://github.com/joelklabo/lnurl-rfc/blob/luds/08.md

PR #243, opened 2023-10-13
LUD-XX: Add Onchain Addresses as optional field in PayRequest response

SachinMeier:


adiko02:
5MpNzCjNWZQjTTmYufvsADEQtKUb3WLTCRLNKJzrBFru

andrerfneves:
relevant: https://github.com/lnurl/luds/pull/303

shocknet-justin:
Why not just use the fallback chain address in bolt11? Better for wallets to adhere to that standard rather than adding another.

PR #248, opened 2023-11-19
 Expiry in `payRequest`

rav3r:
Some services that use hodl/wrapped invoices don't accept very short expiry, it would be nice to request an invoice with a longer expiry.

BTW: I followed LUD-12 as my model for this proposal (by Andre Neves)

### DIFF
--- a/21.md
+++ b/21.md
@@ -1,0 +1,40 @@
+LUD-21: Expiry in `payRequest`.
+=================================
+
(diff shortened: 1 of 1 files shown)


dni:
i like it, why not? @hsjoberg

dni:
maybe we should at some point mention the extensions of lud-7 in https://github.com/lnurl/luds/blob/luds/07.md

PR #249, opened 2023-11-26
Update Auth protocol description, add example values, fix bug.

michaelWuensch:
Hi, 
I am in the process of implementing LNURL Auth for BitBanana. It works and it gives the same linking key as Zeus for the same node.
During the process of implementations I stumbled upon some difficulties, that I now try to overcome with this PR so implementation for other users is easier.

First big thing I noticed is, that both Zeus and Blixt do NOT follow the spec.
In LUD-04 at step 2 it said canonical phrase is hashed one time before signing, but Zeus and Blixt hash it two times. (the called LND endpoint hashes it two times automatically, unless you explicitly tell it to only hash it once, which these wallets did not do.)
I now did the same and adapted the spec accordingly. I am unsure though how other wallets do it.

I further 
- clarified how domain name is converted into bytes.
- removed PrivateKey(...) which looked like a function and felt confusing and unnecessary for me.
- linked LUD 05 and LUD 13 in LUD 4
- added an example with values to easily debug an implementation. I verified these values with the Zeus implementation (https://github.com/ZeusLN/zeus/blob/master/views/LnurlAuth.tsx#L112), BitBanana gives the same results now.

### DIFF
--- a/04.md
+++ b/04.md
@@ -55,6 +55,10 @@ Later, once `LN SERVICE` receives a call at the specified `LNURL-auth` handler, 
 
 However, the second goal is not reachable in practice because there exist different formats of seeds which can't be transferred across all existing wallets. As such, a practical approach is to have recommended ways to derive `linkingKey` for different wallet types.
 
+Please refer to the following LUDs for more details:  
+[LUD-05](05.md): BIP32-based seed generation for auth protocol.  
+[LUD-13](13.md): signMessage-based seed generation for auth protocol. 
+
 
 ## Signature check examples
 
--- a/13.md
+++ b/13.md
@@ -16,8 +16,29 @@ Here we define a canonical phrase to be signed, and from that we will derive the
 In this case neither `hashingKey` nor domain-specific `linkingKey`s can be derived by the path. To overcome this limitation a different scheme is used for this class of wallets:
 
 1. The following canonical phrase is defined: `DO NOT EVER SIGN THIS TEXT WITH YOUR PRIVATE KEYS! IT IS ONLY USED FOR DERIVATION OF LNURL-AUTH HASHING-KEY, DISCLOSING ITS SIGNATURE WILL COMPROMISE YOUR LNURL-AUTH IDENTITY AND MAY LEAD TO LOSS OF FUNDS!`.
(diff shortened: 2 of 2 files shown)


hsjoberg:
Just a short comment before looking into this further, the `single_hash` arg to `signMessage` did definitely not exist back when I implemented LUD-04/13 for Blixt.
In fact, I tried to make them [remove the hashing requirement](https://github.com/lightningnetwork/lnd/pull/4495), before we wrote LUD-13.
So I'm not sure when did double-hashing stuff arrived.

michaelWuensch:
Ah, okay.
As double hashing is the default now I just assumed is has always been that way and they only later added the option to do single hashing. At least I can say that I needed to do the double hashing to get the same result as Zeus, I could not test Blixt as Blixt cannot remote control the same node.

michaelWuensch:
@hsjoberg 
I now looked at the other implementations.
Alby and Balance of satoshi seem to do double hashing as well.
I am not sure about CLAMS. It just calls the signMessage Endpoint from core lightning with the plain message as input, but I couldn't find out how often this is hashed be core lightning before signing.

This means:
Blixt, Zeus, Balance of Satoshi & Alby seem to do double hashing. I will join this for BitBanana so I am compatible and others do not have to change their implementation which is bad UX for their users.


References:
Here is the commit that added double hashing for lnd
https://github.com/lightningnetwork/lnd/commit/02757f67350297811615cc2d7cf7a1fe85a34bb7

Alby:
https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/actions/lnurl/auth.ts#L81
https://github.com/getAlby/lightning-browser-extension/blob/master/src/extension/background-script/connectors/lnd.ts#L334

Balance of Satoshi:
https://github.com/alexbosworth/balanceofsatoshis/blob/master/lnurl/auth.js#L109

Clams:
https://github.com/clams-tech/App/blob/master/src/routes/lnurl/utils.ts#L13

hsjoberg:
Thank you so much for your research @michaelWuensch.
Yeah I would've expected this if lnd had a subtle breaking change like this.
I think we should go ahead with the changes you're proposing in this PR.

hsjoberg (approved):

PR #250, opened 2023-12-01
Add Bipa to README

lsunsi:
We at Bipa support some LUDs and intend to support  some more. Feels right to be on the list!


hsjoberg:
Great! 💯

PR #251, opened 2023-12-01
LUDXX - Add proposal for currencies on payRequest

lsunsi:
### Abstract

This PR proposes an extension to the payRequest specification to allow for denomination of amount in different currencies and negotiating a conversion to this currency. It aims to be backwards compatible while taking a further step on providing lightning wallets and services a common language about exchanges and remittances on foreign currencies.

The PR includes what I expect to be a detailed description with some examples of the proposed extension. It also includes some references to related works that can be competing or complementary proposals to this one. This proposal is inspired by those references.

### Implementations

I've been discussing this proposal inside @bipa-app currently and I aim to implement the most recent version of it and testing with partners before merging this PR.

EDIT:

Current proposal implemented at `lsunsi@bipa.app` if anyone wants to check it out.

### DIFF
--- a/21.md
+++ b/21.md
@@ -31,15 +31,16 @@ type BaseResponse = {
 };
 
 type Currency = {
+  code: string, // Code of the currency, used as an ID for it. E.g.: BRL
   name: string, // Name of the currency. E.g.: Reais
   symbol: string, // Symbol of the currency. E.g.: R$
   displayDecimals: number, // Number of decimal places. E.g.: 2
-  price: number, // Number of currency per BTC. E.g.: 185000
+  multiplier: number, // Number of millisatoshis per smallets unit of currency. E.g.: 5405.405
   convertible?: bool // Whether the currency can be converted into.
 }
 
 type ExtendedResponse = BaseResponse & {
-  currencies: { [key: string]: Currency }
+  currencies: Currency[]
 }
 ```
 
@@ -50,15 +51,16 @@ type ExtendedResponse = BaseResponse & {
   "callback": "https://api.bipa.app/ln/request/invoice/kenu",
   "maxSendable": 1000000000,
   "minSendable": 1000,
-+ "currencies": {
-+   "BRL": {
++ "currencies": [
++   {
++     "code": "BRL",
 +     "name": "Real",
 +     "symbol": "R$",
 +     "displayDecimals": 2,
-+     "price": 185000,
++     "multiplier": 5404.405,
 +     "convertible": true
 +   }
-+ }
++ ]
 }
 ```
 
@@ -66,8 +68,9 @@ type ExtendedResponse = BaseResponse & {
 - The inclusion of a `currency` implies it can be used for denomination of an amount
 - The inclusion of a `convertible currency` implies the `SERVICE` can quote and guarantee a price for given currency
 - The price returned on this response is not guaranteed by the `SERVICE` and is subject to change
-- The `key` of a `currency` will be used as an identifier for the next request
-- The `key` and other information must be according to [ISO-4217](https://en.wikipedia.org/wiki/ISO_4217) for currencies that are included in it
+- The `code` of a `currency` will be used as an identifier for the next request
+- The `code` and other information must be according to [ISO-4217](https://en.wikipedia.org/wiki/ISO_4217) for currencies that are included in it
+- The order of the `currencies` must be interpreted by the `WALLET` as the receiving user preference for a currency
 
 ### Wallet-side second request
 
@@ -89,9 +92,11 @@ In a more particular case, when user is denominating the amount in `CURRENCY`, b
 
 In another particular case, when the user is denominating the amount in millisatoshi, but wants it to be converted to `CURRENCY` on `SERVICE`.
 ```diff
-- <callback><?|&>amount=<milliSatoshi>
-+ <callback><?|&>amount=<milliSatoshi>&convert=<CURRENCY>
+- <callback><?|&>amount=<millisatoshi>
++ <callback><?|&>amount=<millisatoshi>&convert=<CURRENCY>
 ```
+
+Note that the amount provided in all requests is always in integer number denominated in the smallets possible unit of the selected currency. The smallest unit need to be um sync with the `displayDecimals` parameter, so the `WALLET` has all the needed information to receive input and show output properly.
 
 ### Service-side second response
 
@@ -132,21 +137,23 @@ These examples show all the possible uses of this extension by a supporting `WAL
   "metadata": "...",
   "minSendable": 1000,
   "maxSendable": 1000000,
-  "currencies": {
-    "BRL": {
+  "currencies": [
+    {
+      "code": "BRL",
       "name": "Reais",
       "symbol": "R$",
       "displayDecimals": 2,
-      "price": 185000,
+      "multiplier": 5405.405,
       "convertible": true
     },
-    "USDT": {
+    {
+      "code": "USDT",
       "name": "Tether",
       "symbol": "₮",
-      "displayDecimals": 2,
-      "price": 38000
+      "displayDecimals": 6,
+      "multiplier": 26315.789
     }
-  }
+  ]
 }
 ```
 
@@ -162,36 +169,36 @@ These examples show all the possible uses of this extension by a supporting `WAL
 // GET bipa.app/callback?amount=538000&convert=BRL
 {
   "pr": "(invoice of 538 sats)",
-  "converted": 1
+  "converted": 100
 }
 ```
-#### Payer sends 1 BRL worth of BTC as BTC
+#### Payer sends 1 BRL (100 cents of BRL) worth of BTC as BTC
 ```json
-// GET bipa.app/callback?amount=1BRL
+// GET bipa.app/callback?amount=100BRL
 {
   "pr": "(invoice of 538 sats)"
 }
 ```
-#### Payer sends 1 BRL worth of BTC as BRL
+#### Payer sends 1 BRL (100 cents of BRL) worth of BTC as BRL
 ```json
-// GET bipa.app/callback?amount=1BRL&convert=BRL
+// GET bipa.app/callback?amount=100BRL&convert=BRL
 {
   "pr": "(invoice of 538 sats)",
-  "converted": 1
+  "converted": 100
 }
 ```
-#### Payer sends 1 BRL worth of BTC as USDT
+#### Payer sends 1 BRL (100 cents of BRL) worth of BTC as USDT (20 cents of USDT)
 ```json
-// GET bipa.app/callback?amount=1BRL&convert=USDT
+// GET bipa.app/callback?amount=100BRL&convert=USDT
 {
   "pr": "(invoice of 538 sats)",
-  "converted": 0.2
+  "converted": 200000
 }
 ```
-#### Payer sends 1 BRL worth of BTC to unsupported
+#### Payer sends 1 BRL (100 cents of BRL) worth of BTC to unsupported
 ```json
-// GET bipa.app/callback?amount=1BRL
-// ERROR, because 1BRL is not a number and can't
+// GET bipa.app/callback?amount=100BRL
+// ERROR, because 100BRL is not a number and can't
 // be interpret as milisatoshis by the unsupported service.
 ```
 
--- a/21.md
+++ b/21.md
@@ -107,7 +107,16 @@ Note that the amount provided in all requests is always an integer number interp
 
 Upon receiving a currency-denominated request from `WALLET`, the `SERVICE` must return an invoice with an amount matching the converted rate for the amount in that currency. The rate used does not need to match the `multiplier` first informed.
 
-If the `WALLET` requested an actual conversion, the `SERVICE` must provide an additional field alongside the invoice informing the guaranteed `converted` amount that will be creditted to the receiver upon payment. The `converted` amount, and therefore the conversion rate, must be guaranteed by the `SERVICE` for as long as the invoice is not expired. The `converted` amount must be denominated in the smallest unit of the currency, just like the `amount` parameter.
+If the `WALLET` requested an actual conversion, the `SERVICE` must provide an additional `converted` field alongside the invoice informing the guaranteed converted`amount` that will be creditted to the receiver upon payment. The `converted amount`, and therefore the conversion rate, must be guaranteed by the `SERVICE` for as long as the invoice is not expired. The `converted amount` must be denominated in the smallest unit of the currency, just like the `amount` parameter on the callback.
+
+Alongside the `amount` in the `converted` object, the `SERVICE` must also inform how many millisatoshis will be taken as `fee` for the conversion. Finally, a new `multiplier` in the same form as on the first request, must also be present.
+
+The following restriction **must** be met:
+> invoice amount msat = `amount` * `multiplier` + `fee`
+
+This criteria implies the `amount` must be in the smallest unit of currency and net of `fee`.
+The `fee` must be in millisatoshis and should be converted into millisatoshis if taken in other currencies.
+The `multiplier` will act as final price net of fee of the conversion.
 
 ```typescript
 type BaseResponse = {
@@ -116,7 +125,11 @@ type BaseResponse = {
 }
 
 type ExtendedResponse = BaseResponse & {
-  converted?: number, // Integer; Present if and only if `convert` was received.
+  converted?: {
+    multiplier: number, // Double; The quoted multiplier of the conversion.
+    amount: number, // Integer; Number of currency smallest units the payer will receive after fee.
+    fee: number // Integer; Number of millisatoshis representing the fee taken for the conversion.
+  }
 }
 ```
 
@@ -124,7 +137,11 @@ type ExtendedResponse = BaseResponse & {
 {
   "pr": "lnbc1230n1pjknkl...ju36m3lyytlwv42fee8gpt6vd2v",
   "routes": [],
-+ "converted": 123
++ "converted": {
++   "multiplier": 4321.123,
++   "amount": 123,
++   "fee": 1
++ }
 }
 ```
 
@@ -175,12 +192,12 @@ These examples show all the possible uses of this extension by a supporting `WAL
 ###### Payer sends 538 sats to be converted into BRL
 ```json5
 // GET <callback>?amount=538000&convert=BRL
-{ "converted": 100, "pr": "(invoice of 538 sats)" }
+{ "pr": "(invoice of 538 sats)", "converted": { "amount": 100, "fee": 1000, "multiplier": 5370 } }
 ```
 ###### Payer sends 1 BRL worth of BTC to be converted into USDT
 ```json5
 // GET <callback>?amount=100.BRL&convert=USDT
-{ "converted": 200000, "pr": "(invoice of 538 sats)" }
+{ "pr": "(invoice of 538 sats)", "converted": { "amount": 200000, "fee": 2000, "multiplier": 2.68 } }
 ```
 ###### Payer sends 1 BRL worth of BTC to unsupported service
 ```json5
--- a/21.md
+++ b/21.md
@@ -41,7 +41,10 @@ type Currency = {
   symbol: string, // Symbol of the currency. E.g.: R$
   decimals: number, // Integer; Number of decimal places. E.g.: 2
   multiplier: number, // Double; Number of millisatoshis per smallest unit of currency. E.g.: 5405.405
-  convertible?: bool // Whether the payment can be converted into the currency
+  convertible?: { // Whether the payment can be converted into the currency
+    max: number, // Integer; Max converted amount of currency
+    min: number // Integer; Min converted amount of currency
+  }
 }
 
 type ExtendedResponse = BaseResponse & {
@@ -63,7 +66,10 @@ type ExtendedResponse = BaseResponse & {
 +     "symbol": "R$",
 +     "decimals": 2,
 +     "multiplier": 5404.405,
-+     "convertible": true
++     "convertible": {
++       "max": 100000,
++       "min": 1000
++     }
 +   }
 + ]
 }
@@ -71,11 +77,12 @@ type ExtendedResponse = BaseResponse & {
 
 - The inclusion of the `currencies` field implies the support of this extension
 - The inclusion of a `currency` implies it can be used for denomination of an amount
-- The inclusion of a `convertible currency` implies the `SERVICE` can quote and guarantee a price for a given currency
+- The inclusion of a `convertible` field implies the `SERVICE` can quote and guarantee a price for a given `currency`
 - The `multiplier` is not guaranteed by the `SERVICE` and is subject to change
 - The `code` of a `currency` will be used as an identifier for the next request and must be unique
 - The `code` must be according to [ISO-4217](https://en.wikipedia.org/wiki/ISO_4217) if possible
 - The order of the `currencies` may be interpreted by the `WALLET` as the receiving user preference for a currency
+- The `max` and `min` fields within `convertible` field must be respected by `WALLET` on `convertible` requests
 
 ### Wallet-side second request
 
@@ -140,7 +147,10 @@ These examples show all the possible uses of this extension by a supporting `WAL
       "symbol": "R$",
       "decimals": 2,
       "multiplier": 5405.405,
-      "convertible": true
+      "convertible": {
+        "max": 100000,
+        "min": 100
+      }
     },
     {
       "code": "USDT",
--- a/21.md
+++ b/21.md
@@ -34,7 +34,7 @@ type Currency = {
   code: string, // Code of the currency, used as an ID for it. E.g.: BRL
   name: string, // Name of the currency. E.g.: Reais
   symbol: string, // Symbol of the currency. E.g.: R$
-  displayDecimals: number, // Number of decimal places. E.g.: 2
+  decimals: number, // Number of decimal places. E.g.: 2
   multiplier: number, // Number of millisatoshis per smallets unit of currency. E.g.: 5405.405
   convertible?: bool // Whether the currency can be converted into.
 }
@@ -56,7 +56,7 @@ type ExtendedResponse = BaseResponse & {
 +     "code": "BRL",
 +     "name": "Real",
 +     "symbol": "R$",
-+     "displayDecimals": 2,
++     "decimals": 2,
 +     "multiplier": 5404.405,
 +     "convertible": true
 +   }
@@ -96,7 +96,7 @@ In another particular case, when the user is denominating the amount in millisat
 + <callback><?|&>amount=<millisatoshi>&convert=<CURRENCY>
 ```
 
-Note that the amount provided in all requests is always an integer number denominated in the smallets possible unit of the selected currency. The smallest unit need to be in sync with the `displayDecimals` parameter, so the `WALLET` has all the needed information to receive input and show output properly.
+Note that the amount provided in all requests is always an integer number denominated in the smallets possible unit of the selected currency. The smallest unit need to be in sync with the `decimals` parameter, so the `WALLET` has all the needed information to receive input and show output properly.
 
 ### Service-side second response
 
@@ -142,7 +142,7 @@ These examples show all the possible uses of this extension by a supporting `WAL
       "code": "BRL",
       "name": "Reais",
       "symbol": "R$",
-      "displayDecimals": 2,
+      "decimals": 2,
       "multiplier": 5405.405,
       "convertible": true
     },
@@ -150,7 +150,7 @@ These examples show all the possible uses of this extension by a supporting `WAL
       "code": "USDT",
       "name": "Tether",
       "symbol": "₮",
-      "displayDecimals": 6,
+      "decimals": 6,
       "multiplier": 26315.789
     }
   ]
--- a/21.md
+++ b/21.md
@@ -13,11 +13,11 @@ This document describes an extension to the [payRequest](https://github.com/lnur
 
 The main features provided by this extension are:
 - `SERVICE` **MUST** inform `WALLET` what currencies it supports
-- `WALLET` **MAY** request an invoice with amount denominated in one of the currencies
+- `WALLET` **MAY** request an invoice with an amount denominated in one of the currencies
 - `WALLET` **MAY** request to the payment to be converted into one of the currencies
-- `WALLET` **MAY** inform the `SERVICE` the amount and currency used the buy the amount being paid
+- `WALLET` **MAY** inform the `SERVICE` of the amount and currency used to buy the amount being paid
 
-The extension is opt-in and backwards compatible. Further, a supporting `WALLET` can always tell if a `SERVICE` is also supporting beforehand so the communication is never ambiguous.
+The extension is opt-in and backward compatible. Further, a supporting `WALLET` can always tell if a `SERVICE` is also supporting beforehand so the communication is never ambiguous.
 
 ### Wallet-side first request
 
@@ -25,7 +25,7 @@ The first request is unchanged from the base specification.
 
 ### Service-side first response
 
-`SERVICE` must alter its JSON response to the first callback to include a `currencies` field, as follows:
+`SERVICE` must alter its JSON response to the first request to include a `currencies` field, as follows:
 
 ```typescript
 type BaseResponse = {
@@ -41,8 +41,8 @@ type Currency = {
   name: string, // Name of the currency. E.g.: Reais
   symbol: string, // Symbol of the currency. E.g.: R$
   decimals: number, // Integer; Number of decimal places. E.g.: 2
-  multiplier: number, // Double; Number of millisatoshis per smallets unit of currency. E.g.: 5405.405
-  convertible?: bool // Whether the currency can be converted into.
+  multiplier: number, // Double; Number of millisatoshis per smallest unit of currency. E.g.: 5405.405
+  convertible?: bool // Whether the payment can be converted into the currency
 }
 
 type ExtendedResponse = BaseResponse & {
@@ -72,15 +72,15 @@ type ExtendedResponse = BaseResponse & {
 
 - The inclusion of the `currencies` field implies the support of this extension
 - The inclusion of a `currency` implies it can be used for denomination of an amount
-- The inclusion of a `convertible currency` implies the `SERVICE` can quote and guarantee a price for given currency
-- The `multiplier` returned on this response is not guaranteed by the `SERVICE` and is subject to change
+- The inclusion of a `convertible currency` implies the `SERVICE` can quote and guarantee a price for a given currency
+- The `multiplier` is not guaranteed by the `SERVICE` and is subject to change
 - The `code` of a `currency` will be used as an identifier for the next request and must be unique
-- The `code` and other information must be according to [ISO-4217](https://en.wikipedia.org/wiki/ISO_4217) for currencies that are included in it
-- The order of the `currencies` must be interpreted by the `WALLET` as the receiving user preference for a currency
+- The `code` must be according to [ISO-4217](https://en.wikipedia.org/wiki/ISO_4217) if possible
+- The order of the `currencies` may be interpreted by the `WALLET` as the receiving user preference for a currency
 
 ### Wallet-side second request
 
-Upon receiving the `currencies` field on the response, the `WALLET` shows the user it has the option of denominating the amount in one of the `currencies` or for the payment to be credit as the different currency for the receiver. If the BTC being used for the payment is was bought as part of the payment, the `WALLET` may inform the `SERVICE` about it's cost.
+Upon receiving the `currencies` field on the response, the `WALLET` shows the user it has the option of denominating the amount in one of the `currencies` or for the payment to be creditted as a different `currency` for the receiver. If the BTC being used for the payment is was bought as part of the payment, the `WALLET` may inform the `SERVICE` about it's cost.
 
 The inputs that must be gathered from the user are:
 - An optional denominating currency and amount (`CURRENCY_D` and `AMOUNT_D`)
@@ -88,22 +88,22 @@ The inputs that must be gathered from the user are:
 - An optional target currency (`CURRENCY_T`)
 
 The most general case has all the possible parameters present.
-It will generate an invoice with amount equivalent to `AMOUNT_D` `CURRENCY_D`, that will be converted into `CURRENCY_T` by the `SERVICE` upon payment, which will be done with BTC that costed the sender `AMOUNT_S` `CURRENCY_S`.
+It will generate an invoice with the amount equivalent to `AMOUNT_D` `CURRENCY_D`, which will be converted into `CURRENCY_T` by the `SERVICE` upon payment, which will be done with BTC that costs the sender `AMOUNT_S` `CURRENCY_S`.
 
 `<callback><?|&>amount=<AMOUNT_D>.<CURRENCY_D>&source=<AMOUNT_S>.<CURRENCY_S>&target=<CURRENCY_T>`
 
 Each combination of parameters is valid and generates a different use case.
-- Ommiting the `amount` denomination implies the invoice is for millisatoshis (base spec)
-- Ommiting the `source` prevents the receiver from knowing the cost of the BTC being sent
-- Ommiting the `target` implies the receiver will get BTC from the payment, no matter the `source` or `amount` denomination
+- Omitting the `amount` denomination implies the invoice is for millisatoshis (base spec)
+- Omitting the `source` prevents the receiver from knowing the cost of the BTC being sent
+- Omitting the `target` implies the receiver will get BTC from the payment, no matter the `source` or `amount` denomination
 
 Note that the amount provided in all requests is always an integer number interpreted as the smallest unit of the selected `currency`. The smallest unit needs to be according to the `decimals` parameter, so the `WALLET` has all the needed information to receive input and show output properly.
 
 ### Service-side second response
 
-Upon receiving a currency denominated request from `WALLET`, the `SERVICE` must return an invoice with an amount matching the converted rate for the amount in that currency. The rate used does not need to match the `multiplier` first informed.
+Upon receiving a currency-denominated request from `WALLET`, the `SERVICE` must return an invoice with an amount matching the converted rate for the amount in that currency. The rate used does not need to match the `multiplier` first informed.
 
-If the `WALLET` requested an actual conversion, the `SERVICE` must provide an additional field alongside the invoice informing the guaranteed `converted` amount that will be credit to the receiver upon payment. The `converted` amount, and therefore the conversion rate, must be guaranteed by the `SERVICE` for as long as the invoice is not expired. The `converted` amount must be denominated in the smallest unit of the currency, just like the `amount` parameter.
+If the `WALLET` requested an actual conversion, the `SERVICE` must provide an additional field alongside the invoice informing the guaranteed `converted` amount that will be creditted to the receiver upon payment. The `converted` amount, and therefore the conversion rate, must be guaranteed by the `SERVICE` for as long as the invoice is not expired. The `converted` amount must be denominated in the smallest unit of the currency, just like the `amount` parameter.
 
 If the `WALLET` informed the `SERVICE` about the cost of the BTC used in the payment, the `SERVICE` may use the information to display currency-to-currency exchange rates.
 
@@ -172,7 +172,7 @@ These examples show all the possible uses of this extension by a supporting `WAL
 // GET <callback>?amount=538000&target=BRL
 { "converted": 100, "pr": "(invoice of 538 sats)" }
 ```
-###### Payer sends 538 sats that costed 0.2 USDT
+###### Payer sends 538 sats that cost 0.2 USDT
 ```json5
 // GET <callback>?amount=538000&source=200000.USDT
 { "pr": "(invoice of 538 sats)" }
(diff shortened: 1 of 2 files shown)


jklein24:
Nice :-D. As discussed on telegram, it would be great to align this with the existing LUD-21 proposal (and UMA) where possible. UMA actually already extended that proposal in a few of the ways you've done here:

1) Using multiple currencies in the first response.
2) Locking in the exchange rate implicitly via invoice expiration time (see https://arc.net/l/quote/nbilkwwa)
3) Including `displayDecimals` in the Currency object (would be nice to align on naming).

A few super simple changes to make these more consistent:

1) Change the name and semantics of `price` to match `multiplier` in the existing spec
2) Use `displayDecimals` instead of `decimals`
3) Use an array instead of an object for currencies and add the `code` field to the currency. The main reason for this is that it allows a sense of ordering for preferences by the payee. If I prefer BRL, but can also receive BTC, I'm able to set that preference via the ordering within the array.
4) Optional: include the max and min sendable fields from the other proposal.

Also a very minor spelling nit - convertable -> convertible.

21.md
```diff
@@ -0,0 +1,202 @@
...
+    maxSendable: number,
+    minSendable: number
+};
+
+type Currency = {
+  name: string, // Name of the currency. E.g.: Reais
+  symbol: string, // Symbol of the currency. E.g.: R$
+  decimals: number, // Number of decimal places. E.g.: 2
+  price: number, // Number of currency per BTC. E.g.: 185000
+  convertable?: bool // Whether the currency can be converted into.
```
jklein24: Spelling nit: 
```suggestion
  convertible?: bool // Whether the currency can be converted into.
```

21.md
```diff
@@ -0,0 +1,202 @@
...
+  "callback": "https://api.bipa.app/ln/request/invoice/kenu",
+  "maxSendable": 1000000000,
+  "minSendable": 1000,
++ "currencies": {
++   "BRL": {
++     "name": "Real",
++     "symbol": "R$",
++     "decimals": 2,
++     "price": 185000,
++     "convertable": true
```
jklein24:
```suggestion
+     "convertible": true
```

21.md
```diff
@@ -0,0 +1,202 @@
...
++   }
++ }
+}
+```
+
+- The inclusion of the `currencies` field implies the support of this extension
+- The inclusion of a `currency` implies it can be used for denomination of an amount
+- The inclusion of a `convertable currency` implies the `SERVICE` can quote and guarantee a price for given currency
```
jklein24:
```suggestion
- The inclusion of a `convertible currency` implies the `SERVICE` can quote and guarantee a price for given currency
```

21.md
```diff
@@ -0,0 +1,202 @@
...
+  "metadata": "...",
+  "minSendable": 1000,
+  "maxSendable": 1000000,
+  "currencies": {
+    "BRL": {
+      "name": "Reais",
+      "symbol": "R$",
+      "decimals": 2,
+      "price": 185000,
+      "convertable": true
```
jklein24:
```suggestion
      "convertible": true
```

21.md
```diff
@@ -0,0 +1,202 @@
...
+- The `key` of a `currency` will be used as an identifier for the next request
+- The `key` and other information must be according to [ISO-4217](https://en.wikipedia.org/wiki/ISO_4217) for currencies that are included in it
+
+### Wallet-side second request
+
+Upon seeing the `currencies` field on the response from `SERVICE`, a `WALLET` may show the user it has the option of denominating the amount in different currencies or even of sending sats to be credited as the different currency for the receiver.
+
+After gathering the input from the user, it must be encoded into the callback as query parameters as follows.
+
+The most general case, when the user is denominating the amount in `CURRENCY1`, but wants it to be converted to `CURRENCY2` on `SERVICE`.
```
jklein24: I'm slightly confused on who's responsible for knowing the conversions here. Is it always the payee service that needs to know both CURRENCY1 and CURRENCY2 and their exchange rates to/from bitcoin?

This differs slightly from the UMA flow (which may be fine since it's a slightly different use case). With UMA , the sender's service (wallet/custodian/exchange/etc) would be responsible for converting from the sending currency to bitcoin before traveling over lightning. For that reason, its communication of its fees and conversion rate to the user are outside of the scope of UMA/LNURL and just between it and its user. It's really just always bitcoin being sent to the receiver and the payee service doesn't need to care what currency it was before it was bitcoin.

lsunsi: You understood it correctly, but let me try to clarify. If I ask for a lightning invoice of 1BRL, the payee service will have know the conversion rate and return me with an invoice of about 538 sats. Then if I ask the payment to covnerted into BRL, then the payee service (again) will be responsible for performing the exchange and crediting the payee with BRL.

So the payer never has to know any exchange information, it's all on the payee side. It might want to check exchange rate or something, but it's optional and out of scope of the proposal.

About your second point, I'm actually really glad you mentioned it, because what I call colloquially "source exchange data" was part of the [first draft](https://github.com/bipa-app/lnurl-rfc/pull/1) I thought about this subject. In order to really "simulate" a remittance, we need the payee service to be able to tell which currency the BTC came from and at what cost.

I decided not to add this part to this spec because I think it can be thought of a separate complementary LUD, which as soon as this one is done I'll work on immediately. So at this point, the 'source exchange data' is outside the scope of this proposal and LNURL, like you said, and bitcoin is the only thing being sent and the payee service doesn't know anything about the payer sourcing of bitcoin.

Does that clear things out? I'm happy to keep discussing if things are not clear

21.md
lsunsi: Thanks!

21.md
lsunsi: Thanks!

21.md
lsunsi: Thanks!

21.md
lsunsi: Thanks!

lsunsi:
@jklein24 Some comments on your 

> Change the name and semantics of price to match multiplier in the existing spec

I have no problems making this change, but would you explain to me why it's better? It feels to me like `multiplier` makes sense on the context of @ethanrose proposal, but in mine it can even hinder understanding (since you have to multiply it if you input the value in currency, but you have to divide by it in case you input in BTC). Again, I'm just trying to understand better, because readon 38000 in this field might be easier to understand than 2500000 (in case of USD, 

> Use displayDecimals instead of decimals

Changed!

> Use an array instead of an object for currencies and add the code field to the currency. The main reason for this is that it allows a sense of ordering for preferences by the payee. If I prefer BRL, but can also receive BTC, I'm able to set that preference via the ordering within the array.

This is really interesting. Thinking about a little, how would you say the preference is BTC if BTC is not listed in the currencies? Moreover, maybe this could be achieved with a separate preferredCurrency field or something like that? I really like this idea, just thinking of the best way to achieve it.

> Optional: include the max and min sendable fields from the other proposal.

I thought about that as well, but thought that maybe it could be an extension of the proposal, since the max/min already present kind of fulfills this role? What do you think? I'd love to hear what @ethanrose thinks as well.

> Also a very minor spelling nit - convertable -> convertible.

Not very minor, cmon, I'd hate to merge a spelling mistake into this repo haha thanks! fixed

jklein24:
> This is really interesting. Thinking about a little, how would you say the preference is BTC if BTC is not listed in the currencies? Moreover, maybe this could be achieved with a separate preferredCurrency field or something like that? I really like this idea, just thinking of the best way to achieve it.

The order of currencies in the array dictates order of preferences between currencies. So if I prefer BRL, but can also get BTC, you'd have your currencies array contain BRL first and then BTC. You can check out the [UMA protocol](https://github.com/uma-universal-money-address/protocol/blob/main/umad-04-lnurlp-response.md) to see the full structure, but essentially it would look like:

```
"currencies": [
    {
      "code": "BRL",
      "name": "Brazilian Real",
      "symbol": "R$",
      "minSendable": 1,
      "maxSendable": 1_000_000,
      // Estimated millisats per "unit" (eg. 1 cent in USD)
      "multiplier": 489,
      // Number of digits after the decimal point for display on the sender side. For example,
      // in USD, by convention, there are 2 digits for cents - $5.95. in this case, `displayDecimals`
      // would be 2. Note that the multiplier is still always in the smallest unit (cents). This field
      // is only for display purposes. The sender should assume zero if this field is omitted, unless
      // they know the proper display format of the target currency.
      "displayDecimals": 2,
    },
    {
      "code": "SAT",
      "name": "Satoshis",
      "symbol": "sat",
      "minSendable": 1,
      "maxSendable": 10_000_000,
      "multiplier": 1_000, // estimated millisats per "unit" (eg. 1 cent in USD)
      "displayDecimals": 0,
    },
  ],
```

jklein24:
> I have no problems making this change, but would you explain to me why it's better? It feels to me like multiplier makes sense on the context of @ethanrose proposal, but in mine it can even hinder understanding (since you have to multiply it if you input the value in currency, but you have to divide by it in case you input in BTC). Again, I'm just trying to understand better, because reading 38000 in this field might be easier to understand than 2500000 (in case of USD,

Would love @ethanrose's take on this too. Tbh I wasn't thinking it was better for some technical reason, but more just because so many entities are already live with the `multiplier` name, so staying consistent would be nice. It sounds like maybe I'm still missing some difference between `price` and `multiplier` though (besides just the exponent). I wasn't thinking you'd need to change whether you multiply or divide based on the input currency, but rather that it's just always the multipler from your input currency to mSats. What am I missing there?

lsunsi:
@jklein24 Maybe I am the one missing something!

Let me try to explain.
Considering the `price` is 200000 and `multiplier` is 500000, consider this example.
 
###### First example
The user inputs that it wants to send 1 BRL to the other side.
The request equivalent of this input would be `GET bipa.app/callback?amount=1BRL&convert=BRL`.

If the `WALLET` wants to preview (before calling the callback) what the amount of the invoice would be, what should it do?
If it has a `price`, it needs to `approximate amount = 1 / price * 1e3` in  millisatoshis.
If it has a `multiplier`, it needs to `approximate amount = multiplier * 1` in millisatoshis.

This is the use-case supported by @ethanrose PR and it looks nicer having the multiplier in this case.

###### Second example
The user inputs that it wants to send 500 sats as BRL to the other side.
The request equivalent of this input`GET bipa.app/callback?amount=500000&convert=BRL`

If the `WALLET` wants to preview (before calling the callback) what the amount reaching the destination, what should it do?
If it has a `price`, it needs to `approximate amount = 500 / 1e8  * price` in  BRL.
If it has a `multiplier`, it needs to `approximate amount = 500 * 1e3 / multiplier` in BRL.

###### Conclusion

In the second case you use the multiplier as the denominator, which I think can be confusing this it's called "multiplier".
Both cases work of course, because price and multiplier are equivalent, my only question is about the name "multiplier" being confusing, but maybe I'm the only one getting confused haha.

Does these examples help in any way clear up the confusion?
Also, since you first proposed the `multiplier` field, maybe you want to weight in on what do you prefer for this proposal too @fiatjaf

21.md
jklein24: Nice, I'm glad to hear you've already thought a bit about the source exchange data! Having it as separate to this proposal sounds fine to me as long as it's extensible enough to add that ability in the future.

> In order to really "simulate" a remittance, we need the payee service to be able to tell which currency the BTC came from and at what cost.

Do you mind explaining why the payee service needs to know this? I don't have a ton of knowledge about traditional remittences, but I'd love to learn :-).

jklein24:
> @jklein24 Maybe I am the one missing something!
> 
> Let me try to explain. Considering the `price` is 200000 and `multiplier` is 500000, consider this example.
> 
> ###### First example
> The user inputs that it wants to send 1 BRL to the other side. The request equivalent of this input would be `GET bipa.app/callback?amount=1BRL&convert=BRL`.
> 
> If the `WALLET` wants to preview (before calling the callback) what the amount of the invoice would be, what should it do? If it has a `price`, it needs to `approximate amount = 1 / price * 1e3` in millisatoshis. If it has a `multiplier`, it needs to `approximate amount = multiplier * 1` in millisatoshis.
> 
> This is the use-case supported by @ethanrose PR and it looks nicer having the multiplier in this case.
> 
> ###### Second example
> The user inputs that it wants to send 500 sats as BRL to the other side. The request equivalent of this input`GET bipa.app/callback?amount=500000&convert=BRL`
> 
> If the `WALLET` wants to preview (before calling the callback) what the amount reaching the destination, what should it do? If it has a `price`, it needs to `approximate amount = 500 / 1e8 * price` in BRL. If it has a `multiplier`, it needs to `approximate amount = 500 * 1e3 / multiplier` in BRL.
> 
> ###### Conclusion
> In the second case you use the multiplier as the denominator, which I think can be confusing this it's called "multiplier". Both cases work of course, because price and multiplier are equivalent, my only question is about the name "multiplier" being confusing, but maybe I'm the only one getting confused haha.
> 
> Does these examples help in any way clear up the confusion? Also, since you first proposed the `multiplier` field, maybe you want to weight in on what do you prefer for this proposal too @fiatjaf

Thanks for the writeup - that makes things much more clear and concrete. I guess the thing is that in your two scenarios, you're asking 2 different questions:

- Scenario 1: If the WALLET wants to preview (before calling the callback) **what the amount of the invoice would be**
- Scenario 2: If the WALLET wants to preview (before calling the callback) **what the amount reaching the destination**

To me, it makes sense that these different questions would require some different math. In fact for the first scenario, the easiest thing to do would just be to decode the invoice you get back in the `pr` field, rather than doing the conversion again out-of-band.

lsunsi:
@jklein24 I just updated the proposal considering some of the discussed feedback. https://github.com/lnurl/luds/pull/251/commits/55384ff0c35027f67d8305e4dab58d13d61e35a9 Care to take a look?

21.md
```diff
@@ -89,10 +92,12 @@ In a more particular case, when user is denominating the amount in `CURRENCY`, b
 
 In another particular case, when the user is denominating the amount in millisatoshi, but wants it to be converted to `CURRENCY` on `SERVICE`.
 ```diff
-- <callback><?|&>amount=<milliSatoshi>
-+ <callback><?|&>amount=<milliSatoshi>&convert=<CURRENCY>
+- <callback><?|&>amount=<millisatoshi>
++ <callback><?|&>amount=<millisatoshi>&convert=<CURRENCY>
 ```
 
+Note that the amount provided in all requests is always in integer number denominated in the smallets possible unit of the selected currency. The smallest unit need to be um sync with the `displayDecimals` parameter, so the `WALLET` has all the needed information to receive input and show output properly.
```
lorenzolfm: Note that the amount provided in all requests is always *an* integer number denominated

lorenzolfm: smallest*

lorenzolfm: um => in

21.md
```diff
@@ -162,36 +169,36 @@ These examples show all the possible uses of this extension by a supporting `WAL
 // GET bipa.app/callback?amount=538000&convert=BRL
 {
   "pr": "(invoice of 538 sats)",
-  "converted": 1
+  "converted": 100
```
lorenzolfm: should we add that converted is in the smallest currency unit as well?

21.md
lsunsi: ab98b7d6ce1350dd4e2426eb110e7c6cf216c862

21.md
lsunsi: ab98b7d6ce1350dd4e2426eb110e7c6cf216c862

21.md
lsunsi: It's actually so the user can have a nice experience typing (meaning, I can preview how much will it receive or spend while the typing is happening) without actually using the callback (which can be expensive in some implementations). :D

21.md
```diff
@@ -0,0 +1,209 @@
...
++ <callback><?|&>amount=<AMOUNT><CURRENCY>
+```
+
+In another particular case, when the user is denominating the amount in millisatoshi, but wants it to be converted to `CURRENCY` on `SERVICE`.
+```diff
+- <callback><?|&>amount=<millisatoshi>
++ <callback><?|&>amount=<millisatoshi>&convert=<CURRENCY>
+```
+
+Note that the amount provided in all requests is always an integer number denominated in the smallets possible unit of the selected currency. The smallest unit need to be in sync with the `decimals` parameter, so the `WALLET` has all the needed information to receive input and show output properly.
```
jaonoctus: I just want to say that It's really confusing to request 100USDT, intuitively expecting to be `$ 100.00` not `$ 1.00`.
I have no other suggestions in fixing this for now, just my 002 cents

21.md
```diff
@@ -0,0 +1,209 @@
...
+
+Upon receiving a currency denominated request from `WALLET`, the `SERVICE` must return a lightning invoice with an amount matching the converted rate from currency. The rate does not need to be the same as provided on the first response.
+
+If the `WALLET` requested an actual conversion, the `SERVICE` must provide an additional field alongside the invoice informing the guaranteed `converted` amount that will be credit to the receiver upon payment. The `converted` amount, and therefore the conversion rate, must be guaranteed by the `SERVICE` for as long as the lightning invoice is not expired. The `converted` amount must be denominated in the smallest unit of the currency, just like the `amount` parameter.
+
+```typescript
+type BaseResponse = {
+  pr: string,
+  routes: [],
+}
```
jaonoctus: What if I request a non-supported currency? let's say `?amount=100NotSupported`

It should not be the following IMO

```json
{
	"reason": "expired",
	"status": "ERROR"
}
```

21.md
```diff
@@ -0,0 +1,209 @@
...
+    maxSendable: number,
+    minSendable: number
+};
+
+type Currency = {
+  code: string, // Code of the currency, used as an ID for it. E.g.: BRL
+  name: string, // Name of the currency. E.g.: Reais
+  symbol: string, // Symbol of the currency. E.g.: R$
+  decimals: number, // Number of decimal places. E.g.: 2
+  multiplier: number, // Number of millisatoshis per smallets unit of currency. E.g.: 5405.405
```
jaonoctus: When converting the base currency to milisats or sats, should it be rounded away from zero, rounded towards zero or something else?

21.md
lsunsi: Yeah, I agree this is a weak point of the proposal, but I also didn't come with anything better. Things I thought of were: (1) having the currency before the value like USDT100 but it amounts to the same problem; (2) having the currency on another field like amount=100&currency=USDT but again, it doesn't help much and makes the request ambiguous for non-implementers of the LUD.

Just throwing ideas. I'd love to have a less confusing/ambiguous way of expressing the same idea.

> just my 002 cents

haha got me there

21.md
lsunsi: It absolutely should not, it should probably be a more descriptive error of this format, but it can be `SERVICE` dependent, as calling the callback with unsupported currency would be a `WALLET` mistake. Does that make sense? I'll probably change my implementation to return something like ` { "status": "ERROR", "reason": "Currency not supported" }` or something.

21.md
lsunsi: Great question. I think the conversion at this point can be anything (away, towards or anything) because the multiplier works as a volatile hint of the actual conversion rate. The actual guaranteed (for limited time) conversion rate is given by the callback call only. Does that make sense to you?

21.md
lsunsi: If having unambiguous interface isn't a priority, this could be converted to something like `amount={A}&currency={C1}&convert={C2}` or even `amount={A}&input={C1}&convert={C2}`. I gave preference for having the error on non-supporting wallets if you pass the wrong parameters, but it could also be interpreted as client error so we don't need to mix stuff on the amount field.

We discussed this privately too, so feel free to give your 002 cents if you want @jklein24

21.md
jaonoctus: @lsunsi yup, makes lots of sense. thanks for clarifying

lorenzolfm:
Should we add a separator between `amount` and `currency`? It makes parsing a lot easier

21.md
```diff
@@ -0,0 +1,209 @@
...
+The most general case, when the user is denominating the amount in `CURRENCY1`, but wants it to be converted to `CURRENCY2` on `SERVICE`.
+`CURRENCY1` and `CURRENCY2` can be the same, which would the common use case for remittances.
+```diff
+- <callback><?|&>amount=<milliSatoshi>
++ <callback><?|&>amount=<AMOUNT><CURRENCY1>&convert=<CURRENCY2>
+
+In a more particular case, when user is denominating the amount in `CURRENCY`, but does not want to convert it or the service does not support conversion.
+```diff
+- <callback><?|&>amount=<milliSatoshi>
++ <callback><?|&>amount=<AMOUNT><CURRENCY>
```
lorenzolfm: Maybe we should add a separator between `<AMOUNT><CURRENCY>`? This is an example of parsing implemented in Rust using the current spec:

```rust
fn parse_amount_and_maybe_currency(input: &str) -> anyhow::Result<(u64, Option<&str>)> {
    let index = input.find(char::is_alphabetic);

    Ok(if let Some(index) = index {
        let amount = input[..index].parse::<u64>()?;
        let currency = &input[index..];

        (amount, Some(currency))
    } else {
        let amount = input.parse::<u64>()?;

        (amount, None)
    })
}
```

Using a separator would make parsing a lot easier

jaonoctus: @lorenzolfm agreed. I would suggest to  pick one of `-._~`, since they are not encoded in query strings.

See also [RFC3986](https://www.rfc-editor.org/rfc/rfc3986) and [RFC7230](https://datatracker.ietf.org/doc/html/rfc7230#section-2.7.1)

21.md
lsunsi: @jaonoctus Now I'm getting `{"reason":"unsuported currency","status":"ERROR"}` on my implementation if the currency is not supported. It's an improvement

21.md
lsunsi: I agree as well, definitely way better. Here's the change in the proposal: https://github.com/lnurl/luds/pull/251/commits/7bf8f3afadc4b6b4e2ddd0f6b0acb2a4871b7e49

@joelklabo I almost naively picked `:`. Thanks for the suggestion! Dot looks good

21.md
```diff
@@ -0,0 +1,210 @@
...
+
+---
+
+## Support for LNURL-pay currencies
+
+This document describes an extension to the [payRequest](https://github.com/lnurl/luds/blob/luds/06.md) base specification that allows the `WALLET` to send money to a `SERVICE` denominating the amount in a different currency. The features proposed enable many use cases ranging from denominating an invoice in a foreign currency to a remittance-like experience.
+
+The main features provided by this extension are:
+- `SERVICE` **MUST** inform `WALLET` what currencies it supports
+- `WALLET` **MAY** request an invoice with amount denominated in one of the currencies
```
lorenzolfm:
```suggestion
- `WALLET` **MAY** request an invoice with an amount denominated in one of the currencies
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+- `WALLET` **MAY** request to the payment to be converted into one of the currencies
```
lorenzolfm:
```suggestion
- `WALLET` **MAY** request the payment to be converted into one of the currencies
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+- `WALLET` **MAY** inform the `SERVICE` the amount and currency used the buy the amount being paid
```
lorenzolfm:
```suggestion
- `WALLET` **MAY** inform the `SERVICE` of the amount and currency used to buy the amount being paid
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+
+The extension is opt-in and backwards compatible. Further, a supporting `WALLET` can always tell if a `SERVICE` is also supporting beforehand so the communication is never ambiguous.
```
lorenzolfm:
```suggestion
The extension is opt-in and backward-compatible. Further, a supporting `WALLET` can always tell if a `SERVICE` is also supporting beforehand so the communication is never ambiguous.
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+    maxSendable: number,
+    minSendable: number
+}
+
+type Currency = {
+  code: string, // Code of the currency, used as an ID for it. E.g.: BRL
+  name: string, // Name of the currency. E.g.: Reais
+  symbol: string, // Symbol of the currency. E.g.: R$
+  decimals: number, // Integer; Number of decimal places. E.g.: 2
+  multiplier: number, // Double; Number of millisatoshis per smallets unit of currency. E.g.: 5405.405
```
lorenzolfm:
```suggestion
  multiplier: number, // Double; Number of millisatoshis per smallest unit of currency. E.g.: 5405.405
```

21.md
```diff
@@ -0,0 +1,210 @@
...
++     "multiplier": 5404.405,
++     "convertible": true
++   }
++ ]
+}
+```
+
+- The inclusion of the `currencies` field implies the support of this extension
+- The inclusion of a `currency` implies it can be used for denomination of an amount
+- The inclusion of a `convertible currency` implies the `SERVICE` can quote and guarantee a price for given currency
```
lorenzolfm:
```suggestion
- `convertible: true` implies the `SERVICE` can guarantee a price and execute the order upon receiving the payment
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+- The `multiplier` returned on this response is not guaranteed by the `SERVICE` and is subject to change
+- The `code` of a `currency` will be used as an identifier for the next request and must be unique
+- The `code` and other information must be according to [ISO-4217](https://en.wikipedia.org/wiki/ISO_4217) for currencies that are included in it
+- The order of the `currencies` must be interpreted by the `WALLET` as the receiving user preference for a currency
+
+### Wallet-side second request
+
+Upon receiving the `currencies` field on the response, the `WALLET` shows the user it has the option of denominating the amount in one of the `currencies` or for the payment to be credit as the different currency for the receiver. If the BTC being used for the payment is was bought as part of the payment, the `WALLET` may inform the `SERVICE` about it's cost.
```
lorenzolfm:
```suggestion
Upon receiving the `currencies` field on the response, the `WALLET` shows the user it has the option of denominating the amount in one of the `currencies` or for the payment to be credited as a different currency for the receiver. If the BTC being used for the payment was bought as part of the payment, the `WALLET` may inform the `SERVICE` about its cost.
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+
+The inputs that must be gathered from the user are:
+- An optional denominating currency and amount (`CURRENCY_D` and `AMOUNT_D`)
+- An optional source currency and amount (`CURRENCY_S` and `AMOUNT_S`)
+- An optional target currency (`CURRENCY_T`)
+
+The most general case has all the possible parameters present.
+It will generate an invoice with amount equivalent to `AMOUNT_D` `CURRENCY_D`, that will be converted into `CURRENCY_T` by the `SERVICE` upon payment, which will be done with BTC that costed the sender `AMOUNT_S` `CURRENCY_S`.
```
lorenzolfm:
```suggestion
It will generate an invoice with the amount equivalent to `AMOUNT_D` `CURRENCY_D`, which will be converted into `CURRENCY_T` by the `SERVICE` upon payment, which will be done with BTC that costs the sender `AMOUNT_S` `CURRENCY_S`.
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+
+`<callback><?|&>amount=<AMOUNT_D>.<CURRENCY_D>&source=<AMOUNT_S>.<CURRENCY_S>&target=<CURRENCY_T>`
+
+Each combination of parameters is valid and generates a different use case.
+- Ommiting the `amount` denomination implies the invoice is for millisatoshis (base spec)
```
lorenzolfm:
```suggestion
- Omitting the `amount` denomination implies the invoice is for millisatoshis (base spec)
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+- Ommiting the `source` prevents the receiver from knowing the cost of the BTC being sent
```
lorenzolfm:
```suggestion
- Omitting the `source` prevents the receiver from knowing the cost of the BTC being sent
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+- Ommiting the `target` implies the receiver will get BTC from the payment, no matter the `source` or `amount` denomination
```
lorenzolfm:
```suggestion
- Omitting the `target` implies the receiver will get BTC from the payment, no matter the `source` or `amount` denomination
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+
+Note that the amount provided in all requests is always an integer number interpreted as the smallest unit of the selected `currency`. The smallest unit needs to be according to the `decimals` parameter, so the `WALLET` has all the needed information to receive input and show output properly.
+
+### Service-side second response
+
+Upon receiving a currency denominated request from `WALLET`, the `SERVICE` must return an invoice with an amount matching the converted rate for the amount in that currency. The rate used does not need to match the `multiplier` first informed.
```
lorenzolfm:
```suggestion
Upon receiving a currency-denominated request from `WALLET`, the `SERVICE` must return an invoice with an amount matching the converted rate for the amount in that currency. The rate used does not need to match the `multiplier` first informed.
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+
+If the `WALLET` requested an actual conversion, the `SERVICE` must provide an additional field alongside the invoice informing the guaranteed `converted` amount that will be credit to the receiver upon payment. The `converted` amount, and therefore the conversion rate, must be guaranteed by the `SERVICE` for as long as the invoice is not expired. The `converted` amount must be denominated in the smallest unit of the currency, just like the `amount` parameter.
```
lorenzolfm:
```suggestion
If the `WALLET` requested an actual conversion, the `SERVICE` must provide an additional field alongside the invoice informing the guaranteed `converted` amount that will be credited to the receiver upon payment. The `converted` amount, and therefore the conversion rate, must be guaranteed by the `SERVICE` for as long as the invoice has not expired. The `converted` amount must be denominated in the smallest unit of the currency, just like the `amount` parameter.
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+```json5
+// GET <callback>?amount=100.BRL
+{ "pr": "(invoice of 538 sats)" }
+```
+###### Payer sends 538 sats to be converted into BRL
+```json5
+// GET <callback>?amount=538000&target=BRL
+{ "converted": 100, "pr": "(invoice of 538 sats)" }
+```
+###### Payer sends 538 sats that costed 0.2 USDT
```
lorenzolfm:
```suggestion
###### Payer sends 538 sats that cost 0.2 USDT
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+```json5
+// GET <callback>?amount=538000&source=200000.USDT
+{ "pr": "(invoice of 538 sats)" }
+```
+###### Payer sends 1 BRL worth of BTC to be converted into USDT
+```json5
+// GET <callback>?amount=100.BRL&target=USDT
+{ "converted": 200000, "pr": "(invoice of 538 sats)" }
+```
+###### Payer sends 1 BRL worth of BTC that costed 0.2 USDT
```
lorenzolfm:
```suggestion
###### Payer sends 1 BRL worth of BTC that cost 0.2 USDT
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+```json5
+// GET <callback>?amount=100.BRL&source=200000.USDT
+{ "pr": "(invoice of 538 sats)" }
+```
+###### Payer sends 538 sats that costed 0.2 USDT to be converted into BRL
```
lorenzolfm:
```suggestion
###### Payer sends 538 sats that cost 0.2 USDT to be converted into BRL
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+```json5
+// GET <callback>?amount=538000&source=200000.USDT&target=BRL
+{ "converted": 100, "pr": "(invoice of 538 sats)" }
+```
+###### Payer sends 1 BRL worth of BTC that costed 0.2 USDT to converted into BRL
```
lorenzolfm:
```suggestion
###### Payer sends 1 BRL worth of BTC that costs 0.2 USDT to be converted into BRL
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+```json5
+// GET <callback>?amount=100.BRL
+{ "status": "ERROR", "reason": "..." }
+```
+
+### Related work
+
+- Some of the ideas included in this PR were taken from the implementation and discussion on [this PR](https://github.com/lnurl/luds/pull/207). Most precisely, @ethanrose (author) and @callebtc (contributor).
+
+- Some early ideas for this including some other aspects of it were hashed out (but not pull-requested) in this [earlier draft](https://github.com/bipa-app/lnurl-rfc/pull/1) too. Thanks @luizParreira (author), @joosjager (contributor), @za-kk (contributor).
```
lorenzolfm:
```suggestion
- Some early ideas for this including some other aspects of it were hashed out (but not pull-requested) in this [earlier draft](https://github.com/bipa-app/lnurl-rfc/pull/1) too. Thanks, @luizParreira (author), @joosjager (contributor), @za-kk (contributor).
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+
+The extension is opt-in and backwards compatible. Further, a supporting `WALLET` can always tell if a `SERVICE` is also supporting beforehand so the communication is never ambiguous.
+
+### Wallet-side first request
+
+The first request is unchanged from the base specification.
+
+### Service-side first response
+
+`SERVICE` must alter its JSON response to the first callback to include a `currencies` field, as follows:
```
lorenzolfm:
```suggestion
`SERVICE` must include a `currencies` field in the response, as follows:
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+    minSendable: number
+}
+
+type Currency = {
+  code: string, // Code of the currency, used as an ID for it. E.g.: BRL
+  name: string, // Name of the currency. E.g.: Reais
+  symbol: string, // Symbol of the currency. E.g.: R$
+  decimals: number, // Integer; Number of decimal places. E.g.: 2
+  multiplier: number, // Double; Number of millisatoshis per smallets unit of currency. E.g.: 5405.405
+  convertible?: bool // Whether the currency can be converted into.
```
lorenzolfm:
```suggestion
  convertible?: bool // Whether SERVICE can convert the payment into the currency.
```

21.md
```diff
@@ -0,0 +1,210 @@
...
++     "convertible": true
++   }
++ ]
+}
+```
+
+- The inclusion of the `currencies` field implies the support of this extension
+- The inclusion of a `currency` implies it can be used for denomination of an amount
+- The inclusion of a `convertible currency` implies the `SERVICE` can quote and guarantee a price for given currency
+- The `multiplier` returned on this response is not guaranteed by the `SERVICE` and is subject to change
```
lorenzolfm:
```suggestion
- The `multiplier` is not guaranteed by the `SERVICE` and is subject to change
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+- The `code` of a `currency` will be used as an identifier for the next request and must be unique
+- The `code` and other information must be according to [ISO-4217](https://en.wikipedia.org/wiki/ISO_4217) for currencies that are included in it
```
lorenzolfm:
```suggestion
- The `code` and `decimal` information must be according to [ISO-4217](https://en.wikipedia.org/wiki/ISO_4217)
```

21.md
```diff
@@ -0,0 +1,210 @@
...
+- The order of the `currencies` must be interpreted by the `WALLET` as the receiving user preference for a currency
```
lorenzolfm:
```suggestion
- The order of the `currencies` may be interpreted by the `WALLET` as the receiving user preference for a currency
```

21.md
jklein24: I would note that using a dot could lead to some weird bugs in practice if someone leaves decimals in the amount in a language like JS that doesn't distinguish between floats and ints. If we use something other than a period, it's a bit more explicit in the handling of bad values.

One more unfortunate JS thing here...

```
parseInt("100.BRL") == 100
parseInt("BRL.100") == NaN
```

I'm a bit worried that if we put the unit after the amount, nodejs lnurl servers will still parse it as a valid sats amount.

lsunsi:
@jklein24 @jaonoctus @lorenzolfm @ethanrose 

I just updated the proposal to include all discussed changes and complement the implementation with another field (`source`).
The text was changed to be more succinct and to the point, while the examples were expanded to cover all cases.

I think this is the best version so far and by far.
I'll implement the proposal on @bipa-app right now so we can showcase it and test.

21.md
```diff
@@ -0,0 +1,210 @@
...
+```
+###### Payer sends 538 sats to be converted into BRL
+```json5
+// GET <callback>?amount=538000&target=BRL
+{ "converted": 100, "pr": "(invoice of 538 sats)" }
+```
+###### Payer sends 538 sats that cost 0.2 USDT
+```json5
+// GET <callback>?amount=538000&source=200000.USDT
+{ "pr": "(invoice of 538 sats)" }
```
jklein24: One thing I might be missing here - in what cases would the SERVICE care what the source cost+currency was? How is that information used? Is there some legal reason this is needed for remittences or something?

lsunsi: The SERVICE might care in order to provide a better experience, in this case, more information about the operation. For example, if you are service A sending dollars to service B to be converted into euros.

The sending service (A) is gonna buy the dollars and the user will confirm the quote for selling into euros, so A will know the whole exchange, including the dollar-to-euro conversion rate.

The receiving service (B) of course knows about the sell, but it wouldn't know about the dollars without being informed by A. Having this information exchanged means both sides can say the actual dollar-to-euro exchange rate and show the information to the user.

Two random examples I can think of are, top of my head: you're actually sending money to yourself overseas or something like transferwise and you'd want to have information about the exchange on both sides. Second one would be you're paying someone an agreed upon amount in your local currency, like I need to pay you $5 for this comment. Trust both services implementations, your wallet would know that I paid $5 for the BTC I sent you, no matter how many sats you got on your side.

Does that explain the logic?

jklein24: I see, so it's mostly for UX on the receiver side so that they can see the whole picture, but they don't actually need to use it for anything on their end. That makes sense! So in the "sender exchange" case we were talking about, it might look something like the case below `GET <callback>?amount=538000&source=200000.USDT&target=BRL` where the sender would specify amount in mSATs, maybe say a source cost, and specify a target currency for the receiver side. That makes sense!

lsunsi: That's exactly what I had in mind. Exactly that!

jklein24 (approved):

jaonoctus (approved):
ACK 7bf8f3afadc4b6b4e2ddd0f6b0acb2a4871b7e49 and 2d37cec01769cbd0812aeae06919bac0104f8bc0

lorenzolfm (approved):

21.md
```diff
@@ -0,0 +1,185 @@
...
+    callback: string,
+    maxSendable: number,
+    minSendable: number
+}
+
+type Currency = {
+  code: string, // Code of the currency, used as an ID for it. E.g.: BRL
+  name: string, // Name of the currency. E.g.: Reais
+  symbol: string, // Symbol of the currency. E.g.: R$
+  decimals: number, // Integer; Number of decimal places. E.g.: 2
```
jklein24: nit: It might be good to include a longer description about this field. With UMA, a bunch of folks integrating have been a bit confused by the mechanics around decimals/multiplier. Might be good to lay out some examples and show why decimals is important for clarity.

jklein24: FYI @lsunsi I added some examples and notes on small currency units in the uma spec that might be helpful here too - https://github.com/uma-universal-money-address/protocol/blob/main/umad-04-lnurlp-response.md#currency-examples

21.md
```diff
@@ -0,0 +1,185 @@
...
+type ExtendedResponse = BaseResponse & {
+  converted?: number, // Integer; Present if and only if `convert` was received.
+}
+```
+
+```diff
+{
+  "pr": "lnbc1230n1pjknkl...ju36m3lyytlwv42fee8gpt6vd2v",
+  "routes": [],
++ "converted": 123
```
jklein24: Funny enough, in UMA, I chose not to include the target amount, but did include the multiplier and some other stuff:

```typescript
"paymentInfo": {
    // The currency code of the receiving currency (eg. "USD"). This should match the requested currency in the payreq
    // request.
    "currencyCode": string,
    // Millisats per "unit" of the receiving currency (eg. 1 cent in USD). A double-precision floating point number.
    // In this context, this is just for convenience. The conversion rate is also baked into the invoice amount itself.
    // `invoice amount = amount * multiplier + exchangeFeesMillisatoshi`
    "multiplier": number,
    // Number of digits after the decimal point for the receiving currency. For example, in USD, by convention, there are
    // 2 digits for cents - $5.95. In this case, `decimals` would be 2. This should align with the currency's `decimals`
    // field in the LNURLP response. It is included here for convenience. See [UMAD-04](/uma-04-local-currency.md) for
    // details, edge cases, and examples.
    "decimals": number,
    // The fees charged (in millisats) by the receiving VASP to convert to the target currency.
    // This is separate from the multiplier rate.
    "exchangeFeesMillisatoshi": number
},
```

The fees are useful for a full visual cost breakdown on the sender side. The multiplier and decimals help infer the converted amount from the invoice amount. I do like having the wrapping struct to keep related fields grouped and allow for extension in the future without polluting the top-level structure. Just some food for thought :-). 

UMA spec link: https://github.com/uma-universal-money-address/protocol/blob/main/umad-06-payreq-response.md

lorenzolfm (approved):
ACK https://github.com/lnurl/luds/pull/251/commits/8c312b46fa1f0779d785571ad4b11d94fd587363

lsunsi:
@jklein24 @lorenzolfm @jaonoctus I hacked this webapp to showcase the LUD. It's using just the API described in this PR. https://lnurlp-live.vercel.app/lorenzolfm@bipa.app

21.md
```diff
@@ -0,0 +1,195 @@
...
+```
+
+- The inclusion of the `currencies` field implies the support of this extension
+- The inclusion of a `currency` implies it can be used for denomination of an amount
+- The inclusion of a `convertible` field implies the `SERVICE` can quote and guarantee a price for a given `currency`
+- The `multiplier` is not guaranteed by the `SERVICE` and is subject to change
+- The `code` of a `currency` will be used as an identifier for the next request and must be unique
+- The `code` must be according to [ISO-4217](https://en.wikipedia.org/wiki/ISO_4217) if possible
+- The order of the `currencies` may be interpreted by the `WALLET` as the receiving user preference for a currency
+- The `max` and `min` fields within `convertible` field must be respected by `WALLET` on `convertable` requests
```
jklein24:
```suggestion
- The `max` and `min` fields within `convertible` field must be respected by `WALLET` on `convertible` requests
```

lsunsi: Fixed!

21.md
lsunsi: This makes a lot of sense, I tried complementing the spec with this info and copied your example kind of . What do you think of (https://github.com/lnurl/luds/pull/251/commits/6a5b6e029f7716a6a6caec75b8b2521fef29d148)?

21.md
lsunsi: This is interesting. The reason I chose to keep just the int is to keep the protocol smaller.
The only property in this structure that is not already included is the fee, right?

The reason I didn't included it is that there's multiple ways of taking the fee, right? It could be billed in sats, or in currency (althought it's kind of the same). But the fee could be in the price (spread) or separate. Etc.

What do you think about these possibilities?

21.md
jklein24: Cool, that captures the gist of it! I think the main difference is that UMA has an explicit max of 8 decimals, but keeping it open is fine too. LGTM

21.md
jklein24: Yeah that's fair. I just delimit the fee in mSats always because my assumption is that it's based into the cost of the invoice: `invoice amount = amount * multiplier + exchangeFeesMillisatoshi`. I like having a top-level structure to wrap payment details since I could imagine more fields being added there in the future for scenarios you're describing. With the wrapper, we wouldn't need to pollute the top level more.

lorenzolfm (approved):

jaonoctus:
@lsunsi noice! I wrote a similar interface ¹ to do the samething 😂  that I showed to @lorenzolfm 

[1](https://twitter.com/jaoNoctus/status/1738976941544296695)

lsunsi:
> @lsunsi noice! I wrote a similar interface to do the samething 😂  that I showed to @lorenzolfm 

Yeah my theory is that he is incepting us with his ideas at the same time!

jaonoctus (approved):

jklein24:
FYI this is now live in [UMA v1](https://github.com/uma-universal-money-address/protocol/blob/main/umad-05-payreq-request.md) and VASPs are rolling it out. Xapo, Ripio, and Bitnob are upgraded and several more are on the way.

lsunsi:
I'd like to point out that besides being part of UMA v1 like @jklein24 , we have a working implementation at [bipa](https://bipa.app), it's a shame this spec doesn't get more attention from maintainers. Maybe @hsjoberg or @fiatjaf could share some thoughts here!

lsunsi:
@lorenzolfm @jaonoctus Updated LUD to 22

jaonoctus (approved):

dni:
:eyes:

lsunsi:
@dni :eyes:

dni:
any interest in this proposal left? :D should we roll it up again?

lsunsi:
@dni definitely! I even have a functional demo of receiving through it since back when hahaha.

We should definitely roll it up, but one question, what does rolling it up mean?

jklein24:
This is also now live [with Bringin](https://x.com/prashanthc123/status/1939754639127408874) in Europe in addition to being used for UMA (Nubank, Bitso, Ripio, Coins, Xapo, Zerohash, etc.) and by Bipa as @lsunsi noted above.

lsunsi:
@jklein24 Bringin is actually pretty off-spec, right? Judging from https://api.bringin.xyz/.well-known/lnurlp/prashanth

jklein24:
> @jklein24 Bringin is actually pretty off-spec, right? Judging from https://api.bringin.xyz/.well-known/lnurlp/prashanth

Good catch! I hadn't actually tested their implementation! Will flag to them on twitter :-)

andrerfneves:
@ethanrose @lsunsi @lorenzolfm @Kukks 

so what's the latest here? Is it this #251 PR that we're converging in as the spec and pushing for numbering LUD22? Or is it the #207 PR? Are they compatible? I know there are many companies already doing this in their own ways, but whats the sensible approach here to get this over the line?

lsunsi:
@andrerfneves

I think this proposal is more complex and more powerful than #207 . Use-case wise it's like a superset of it.

Back in the day I talked with @ethanrose about converging on this one, but I'm not sure how he feels about it today since it was so long ago.

@bipa-app implemented the receiving side of this PR, which is working to this day I think (demoed in here https://lnurlp-live.vercel.app/lsunsi@bipa.app). I'm not aware of any other implementation, both sending and receiving sides.

Anything to add @lorenzolfm ?

Kukks:
There are a couple of overlapping proposals. Let me try and go through them

jklein24:
FYI @andrerfneves @Kukks, this is the proposal that is still active in UMA and moving very substantial volume daily. For example, all SoFi remittances from the US go through this spec, among many other UMA implementors. We'd started with https://github.com/lnurl/luds/pull/207, but moved to this spec over a year ago for the increased flexibility and functionality. Namely, it allows us to lock the sending amount in addition to the receiving amount, which is really important to some use cases (remittances, for example).

PR #252, opened 2023-12-15
LUD-XX: Payee Data.

jklein24:
This is the payee-side equivalent of [LUD-18](18.md). It gives the opportunity for the payer to request that the payee provide identitifying information before the payment is made. This allows the sending `WALLET` to show information about the receiver in order to allow the payer to verify that they are paying the correct entity. This also gives the opportunity for the payee to authorize themselves via a challenge-response mechanism to provide some assurance that `SERVICE` has not been compromised.

This document is largely copied from LUD-18 with some minor tweaks and additional context to make it make sense in the opposite direction.

### DIFF
--- a/22.md
+++ b/22.md
@@ -1,0 +1,70 @@
+LUD-22: Payee identity data in `payRequest` protocol.
+================================================
+
+`author: onthedeklein` `discussion: https://t.me/lnurl/36488`
+
+---
+
+This is the payee-side equivalent of [LUD-18](18.md). It gives the opportunity for the payer to request that the payee provide identitifying information before the payment is made. This allows the sending `WALLET` to show information about the receiver in order to allow the payer to verify that they are paying the correct entity. This also gives the opportunity for the payee to authorize themselves via a challenge-response mechanism to provide some assurance that `SERVICE` has not been compromised.
+
+## 1. `payeeData` record
+
+If `WALLET` wants to get one or more types of payee identities/data from `SERVICE` then it MUST alter its JSON request to the second callback to include a `payeeData` query param, as follows (notice that the `payeeData` record below has a bunch of fields only for completion, an actual response will likely contain just a subset of these):
+
+```typescript
+ {
+  "name": { "mandatory": boolean },
+  "pubkey": { "mandatory": boolean },
+  "identifier": { "mandatory": boolean },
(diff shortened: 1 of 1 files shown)


22.md
```diff
@@ -0,0 +1,70 @@
+LUD-22: Payee identity data in `payRequest` protocol.
+================================================
+
+`author: onthedeklein` `discussion: https://t.me/lnurl/36488`
+
+---
+
+This is the payee-side equivalent of [LUD-18](18.md). It gives the opportunity for the payer to request that the payee provide identitifying information before the payment is made. This allows the sending `WALLET` to show information about the receiver in order to allow the payer to verify that they are paying the correct entity. This also gives the opportunity for the payee to authorize themselves via a challenge-response mechanism to provide some assurance that `SERVICE` has not been compromised.
```
jklein24, line 8: > This also gives the opportunity for the payee to authorize themselves via a challenge-response mechanism to provide some assurance that `SERVICE` has not been compromised.

I'm not totally sure I want to mention this since it doesn't actually provide much assurance there except against particular types of attacks, where the sender already knows and remembered the payer's signing public key retrieved from some other mechanism.

22.md
```diff
@@ -0,0 +1,70 @@
...
+
+```typescript
+ {
+  "name": { "mandatory": boolean },
+  "pubkey": { "mandatory": boolean },
+  "identifier": { "mandatory": boolean },
+  "email": { "mandatory": boolean },
+  "auth": {
+     "mandatory": boolean,
+     "k1": string // hex encoded 32 bytes of challenge
```
jklein24, line 22: Considering changing the name k1 here to avoid confusion with LUD-04, since it's serving a slightly different purpose. Seeking opinions.

jklein24:
FYI this is now live in [UMA v1](https://github.com/uma-universal-money-address/protocol/blob/main/umad-05-payreq-request.md) and VASPs are rolling it out. Xapo, Ripio, and Bitnob are upgraded and several more are on the way.

dni:
love it, totally useful for https://github.com/lnurl/luds/pull/301 to identify mints.

i would add nodeInfo, like peers, channels, capacity, pubkey, color. with this wallets can gauge how reliable the underlaying ln node might be.

PR #253, opened 2023-12-17
Added BitBanana to auth specs

michaelWuensch:
BitBanana 0.7.1 was released and now supports LNURL auth.
This updates the readme accordingly.


hsjoberg:
Great! 💯 Presumably you did as discussed in #249?

michaelWuensch:
Yes, I chose double hashing, resulting in the same account like Zeus for example even though it doesn't match the current official specification.

PR #254, opened 2023-12-22
chore: update blink and galoy support

UncleSamtoshi:
- fixed whitespace for `Bipa` supported specs
- added `Blink` to supported specs
- updated 'Galoy` supported specs


andrerfneves:

Issue #259, opened 2024-05-12
LUD-01: Recommendation to add CORS headers for all server responses


adambor:
Some LNURL services are currently inaccessible from web apps (PWA wallets like Mutiny, swap services like boltz.exchange or atomiq.exchange) due to missing [CORS (Cross-Origin Resource Sharing) header](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS). I would suggest adding a recommendation to add a CORS header for all server response in the base LUD-01 spec, so that LNURLs can be accessed from web apps. It's important for the new and existing LNURL service implementations to understand the need for providing such a header to make their service work with customers using web app wallets or other apps that might have a need to access LNURL services from within the browser environment.

kilrau:
:100:

Whenever we had an issue with lnurl resolution in our web app, it was always CORS. Literally always.

andrerfneves:
Closing as #267 addresses this already

PR #267, opened 2024-06-16
Access-Control-Allow-Origin

shocknet-justin:
https://github.com/lnurl/luds/issues/259

### DIFF
--- a/01.md
+++ b/01.md
@@ -37,7 +37,7 @@ Neither status codes or any HTTP Header has any meaning. Servers may use whateve
 
 ## CORS Browser Compatibility
 
-Application endpoints that need to be accessed when using the LNURL must respond to `OPTIONS` requests on that endpoint. In their replies to `GET` and `OPTIONS`, they must set the header `Access-Control-Allow-Origin: *` for LNURL's to work with browser-based wallets, services and other webapps. If they set the `Access-Control-Allow-Methods` header, they must set in such a way that `GET` requests are permitted.
+Application endpoints or Reverse Proxy directives that are accessed when using the LNURL must respond to `OPTIONS` requests on that endpoint. In their replies to `GET` and `OPTIONS`, they must set the header `Access-Control-Allow-Origin: *` for LNURL's to work with browser-based wallets, services and other webapps. If they set the `Access-Control-Allow-Methods` header, they must set in such a way that `GET` requests are permitted.
 
 ## Decoding examples
 
--- a/01.md
+++ b/01.md
@@ -35,6 +35,10 @@ Once `LNURL` is decoded:
 
 Neither status codes or any HTTP Header has any meaning. Servers may use whatever they want. Clients should ignore them (and be careful when using libraries that treat responses differently based on headers and status codes) and just parse the response body as JSON, then interpret it accordingly.
 
+## CORS Browser Compatibility
+
+Reverse proxies or application endpoints must use the CORS Header `Access-Control-Allow-Origin: *` for LNURL's to work with browser-based wallets, services and other webapps. 
+
 ## Decoding examples
 
 In Scala:
--- a/01.md
+++ b/01.md
@@ -37,7 +37,7 @@ Neither status codes or any HTTP Header has any meaning. Servers may use whateve
 
 ## CORS Browser Compatibility
 
-Application endpoints or Reverse Proxy directives that are accessed when using the LNURL must respond to `OPTIONS` requests on that endpoint. In their replies to `GET` and `OPTIONS`, they must set the header `Access-Control-Allow-Origin: *` for LNURL's to work with browser-based wallets, services and other webapps. If they set the `Access-Control-Allow-Methods` header, they must set in such a way that `GET` requests are permitted.
+Application endpoints (or their defined Reverse Proxy directives) accessed when using the LNURL must respond to `GET` having set the header `Access-Control-Allow-Origin: *` for LNURL's to work with browser-based wallets, services and other webapps. If they set the `Access-Control-Allow-Methods` header, they must set in such a way that `GET` requests are permitted.
 
 ## Decoding examples
 
--- a/01.md
+++ b/01.md
@@ -37,7 +37,7 @@ Neither status codes or any HTTP Header has any meaning. Servers may use whateve
 
 ## CORS Browser Compatibility
 
-Reverse proxies or application endpoints must use the CORS Header `Access-Control-Allow-Origin: *` for LNURL's to work with browser-based wallets, services and other webapps. 
+Application endpoints that need to be accessed when using the LNURL must respond to `OPTIONS` requests on that endpoint. In their replies to `GET` and `OPTIONS`, they must set the header `Access-Control-Allow-Origin: *` for LNURL's to work with browser-based wallets, services and other webapps. If they set the `Access-Control-Allow-Methods` header, they must set in such a way that `GET` requests are permitted.
 
 ## Decoding examples
 


fiatjaf (approved):

01.md
```diff
@@ -35,6 +35,10 @@ Once `LNURL` is decoded:
 
 Neither status codes or any HTTP Header has any meaning. Servers may use whatever they want. Clients should ignore them (and be careful when using libraries that treat responses differently based on headers and status codes) and just parse the response body as JSON, then interpret it accordingly.
 
+## CORS Browser Compatibility
+
+Reverse proxies or application endpoints must use the CORS Header `Access-Control-Allow-Origin: *` for LNURL's to work with browser-based wallets, services and other webapps.
```
AaronDewes:
```suggestion
Application endpoints that need to be accessed when using the LNURL must respond to `OPTIONS` requests on that endpoint. In their replies to `GET` and `OPTIONS`, they must set the header `Access-Control-Allow-Origin: *` for LNURL's to work with browser-based wallets, services and other webapps. If they set the `Access-Control-Allow-Methods` header, they must set in such a way that `GET` requests are permitted.
```

AaronDewes (changes requested):

01.md
freeberty: if add `GET`, could be useful mention `OPTIONS` as well

AaronDewes: > if add `GET`, could be useful mention `OPTIONS` as well

Options requests can not be denied with a header, as they are a browser feature. However, servers should respond to options request. I'll edit my suggestion.

Kukks (approved):
ACK when including @AaronDewes suggestion.

01.md
fiatjaf: how about replacing that entire paragraph with "Requests must support CORS"?

AaronDewes: Maybe `Application endpoints that need to be accessed when interacting with an LNURL must set CORS headers in such a way that they are reachable from any web origin.` would be better?

fiatjaf: Yes.

I just think we shouldn't describe what CORS is.

shocknet-justin: These things are typically done at reverse proxies, people know what nginx and caddy are in their stack. To keep the explanation simple list the header snippets.

hsjoberg (approved):

01.md
AaronDewes: > These things are typically done at reverse proxies, people know what nginx and caddy are in their stack. To keep the explanation simple list the header snippets.

Users don't read these specs and proxies are not responsible for implementing it

AaronDewes: Setting the header in a proxy is a workaround, not a solution.

shocknet-justin: What? Its for the service implementing the specs... the service which runs behind a reverse proxy. Who's directly exposing their app? nginx or caddy are used effectively everywhere, and a reverse proxies directive is part of an application stack. They are not separate.

shocknet-justin (commented):
split the diff

01.md
```diff
@@ -35,6 +35,10 @@ Once `LNURL` is decoded:
 
 Neither status codes or any HTTP Header has any meaning. Servers may use whatever they want. Clients should ignore them (and be careful when using libraries that treat responses differently based on headers and status codes) and just parse the response body as JSON, then interpret it accordingly.
 
+## CORS Browser Compatibility
+
+Application endpoints or Reverse Proxy directives that are accessed when using the LNURL must respond to `OPTIONS` requests on that endpoint. In their replies to `GET` and `OPTIONS`, they must set the header `Access-Control-Allow-Origin: *` for LNURL's to work with browser-based wallets, services and other webapps. If they set the `Access-Control-Allow-Methods` header, they must set in such a way that `GET` requests are permitted.
```
AaronDewes:
```suggestion
Application endpoints that are accessed when using the LNURL must respond to `OPTIONS` requests on that endpoint. In their replies to `GET` and `OPTIONS`, they must set the header `Access-Control-Allow-Origin: *` for LNURL's to work with browser-based wallets, services and other webapps. If they set the `Access-Control-Allow-Methods` header, they must set in such a way that `GET` requests are permitted.
```

I still disagree with the reference to reverse proxies. A reverse proxy is something completely separate and controlled by the user or platform, not by the application developer. 

A reverse proxy is not always part of the application stack, and not under the control of an app developer. For example, BTCPay, which supports LNUrl, can be run on StartOS, Umbrel, ... which may already include a reverse proxy.

Requiring this for apps themselves means apps are correct without users externally modifying something about the app's responses.

If we impose this rule on reverse proxies, a lot of people are going to set `Access-Control-Allow-Origin: *` for all routes, which will make apps insecure.

Clearly putting the responsibility on the app developer is much better imo.

AaronDewes (changes requested):

01.md
shocknet-justin: That's a very niche subset of reverse proxies and afaik they function more as gateways than reverse proxies, an umbrel user exposing services to the internet would need to reconfigure the reverse proxy anyway for things like SSL and so on... I also explicitly mentioned the directive for the endpoint, of which there would be many in an box like that for various apps

AaronDewes: Still, these proxies are controlled by the user and not the application developer, who is responsible for implementing the spec.

shocknet-justin: By exposing an umbrel etc to the internet the user is now a service provider, the box is the application including the customized reverse proxy configuration

It's services that need these fixes, not users.

AaronDewes: > By exposing an umbrel etc to the internet the user is now a service provider, the box is the application including the customized reverse proxy configuration
> 
> It's services that need these fixes, not users.

Every other part of the spec is implemented by app developers, not users. All of [these services](https://github.com/lnurl/luds?tab=readme-ov-file#self-hosted) have implemented LNURL. No user configured the LNURL handling in their reverse proxy or has to read the spec. Application & library developers have to know them, however. 

The application **can** do this, and the reverse proxy is not needed for such functionality, so it should be part of the app.

Saying "our app may support LNURL" is incredibly bad as a developer and complicates setup for users unnecessarily.

shocknet-justin: If we're being precise then those are apps, not services

I agree that apps should implement this, but the reality is that many apps are unlikely to be updated for this and are harder for the service provider to update themselves vs. their reverse proxy configuration

AaronDewes: > If we're being precise then those are apps, not services
> 
> I agree that apps should implement this, but the reality is that many apps are unlikely to be updated for this and are harder for the service provider to update themselves vs. their reverse proxy configuration

I think the spec should talk about apps, and then service providers can fix invalid apps if they choose so, but that should be outside of the spec.

shocknet-justin: Apps are geared towards services, services which are run by operators that may be trying to figure out why teir app isn't working in certian cases- including searching on the CORS error from their browser

A reference to app directives gives both sides the information they need, and doesn't imply app developers shouldn't do it directly

AaronDewes: But it may lead app developers to put the responsibility onto users. Not mentioning reverse proxies makes it clear app developers must do it.

shocknet-justin: The responsibility is already on operators of broken services to fix them, a bad app developer or unmaintained app will deviate no matter whats added to the spec and so verbiage is warranted for both configurations

It's also fair for an app to simply document the external requirement just as it might SSL

AaronDewes: > a bad app developer or unmaintained app will deviate no matter whats added to the spec and so verbiage is warranted for both configurations

Then the app is just broken.  The spec is written for correct apps, not for broken apps.

> It's also fair for an app to simply document the external requirement just as it might SSL

CORS is something completely different from SSL. For CORS, an app should always handle it itself, because the app developers know what CORS rules are safe.

hsjoberg:
Why are we adding `OPTIONS` requests to LNURL?
It's an all-covering breaking change mandating all LNURL server implementations to support it.

I suggest we remove this.

AaronDewes:
> Why are we adding `OPTIONS` requests to LNURL? It's an all-covering breaking change mandating all LNURL server implementations to support it.
> 
> I suggest we remove this.

This is necessary for the LNURL to work/be resolved from web browsers.

AaronDewes:
I was mistaken, LNURL resolution is considered a "simple" request that does not require options.

01.md
```diff
@@ -35,6 +35,10 @@ Once `LNURL` is decoded:
 
 Neither status codes or any HTTP Header has any meaning. Servers may use whatever they want. Clients should ignore them (and be careful when using libraries that treat responses differently based on headers and status codes) and just parse the response body as JSON, then interpret it accordingly.
 
+## CORS Browser Compatibility
+
+Application endpoints (or their defined Reverse Proxy directives) accessed when using the LNURL must respond to `GET` having set the header `Access-Control-Allow-Origin: *` for LNURL's to work with browser-based wallets, services and other webapps. If they set the `Access-Control-Allow-Methods` header, they must set in such a way that `GET` requests are permitted.
```
AaronDewes, line 40:
```suggestion
Application endpoints that need to be accessed when interacting with an LNURL must set [CORS headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) in such a way that they are reachable from any web origin.
```

Based on feedback from @hsjoberg and @fiatjaf, maybe just this sentence would be best.

No someone else apart from me and @shocknet-justin should probably also decide if we should include reverse proxies, because we don't seem to be able to agree here.

adambor: I think we should at least keep the `Access-Control-Allow-Origin: *` or/and a link to the [MDN describing CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS), to make it easier & more straightforward for devs to actually implement.

The issues is that incorrectly configured CORS is not immediately noticeable by a dev who uses non webapp based wallet (majority of wallets currently), I am worried that if this is not straightforward to implement (e.g. devs having to research what this CORS even means) they might just skip through it thinking their LN service is working all good.

Maybe it'd make sense to have a simple online tool (webapp) for checking whether LNURLs adhere to the spec (including CORS)

kaloudis:
Long needed guidance. Thanks for writing this.

ACK

01.md
AaronDewes: @adambor I added your suggestion to my comment above.

@kaloudis @Kukks @adambor What's your opinion about mentioning proxies?

hsjoberg: I think we should remove the mention of proxies to keep the text shorter.

Isn't it already obvious for proxies that CORS header would be required? The proxy is really an "application endpoint" here if I understand it correctly.

kaloudis: Feel like we're bikeshedding on the proxy text. On one hand it's not taking up much space, on the other it also feels obvious to me.

I could go either way, but I lean towards keeping it as is.

AaronDewes: Great! Two more people disagreeing 🙃

Kukks: I think it's obvious and an unnecessary implementation detail to mention.

shocknet-justin:
another one... plz merge

https://x.com/shocknet_justin/status/1823134965225128181

dni (approved):
@AaronDewes what do you think, we have a lot of approves and it should be merged, can you resolve your suggested chance, or you still think its important? for me its either way

Mareklazar:
Schváleno

01.md
otech47: looks good as-is. technical detail is helpful. this should be resolved and merged imo

otech47 (approved):
good guidance

@dni I suggest this be merged. @AaronDewes seems unresponsive and his comments are not contentious

AaronDewes:
I missed the previous mentions. I still think reverse proxies should be removed, but it's not extremely important to me.

I'm not a LNURL spec author and not in a position to block merges anyway :smile: 

Just to repeat my main point:

Mentioning reverse proxies in the spec is (in my opinion) basically an invitation for application developers to not fully implement the spec and "outsource" that to users.

01.md
AaronDewes: >  technical detail is helpful.

This is not technical detail, but rather something unrelated to the spec. You can technically implement even more parts of the spec in a reverse proxy, but it doesn't make sense to mention infrastructure around the servers in the LNUrl spec. The spec does not need to explain that a server should accept TCP packets either.

jamesjyyoung1-ship-it:
Beserious40

PR #269, opened 2024-07-05
Add Bare Bitcoin to LUD6, LUD12

torkelrogstad:
Bare Bitcoin is a Norwegian custodial bitcoin brokerage with support for receiving to a `@bb.no` lightning address, as well as sending to LNURL + LN address when withdrawing


hsjoberg:
Nice 💯

PR #274, opened 2024-12-26
support empty domain.

kehiy:
it can support empty names like nostr nip-05 convention. it looked better for someone who has their domain and they don't have multiple usernames. or a provider can use it as their address. also, it helps people on nostr to have their ln address the same as their nip-05 identifier.


realbub:
I was going to suggest the same. Payment to the domain is a nice addition to LUD-16. Payments to the domain without a username are very useful and may have many use cases like donations to websites, organizations, or payments with a mandatory identifier so the service can differentiate the payer. Also, as mentioned, it is already done in Nostr NIP-05.

PR #277, opened 2025-06-09
feat(LUD-16): support for multiple identifiers

### LINKED ISSUE #197: Support of tags in LUD-16 static identifiers
The idea here is that, additionally to the `<username>@<domainname>` format of [LUD-16 static internet identifiers](https://github.com/lnurl/luds/blob/luds/16.md), a `SERVICE` can offer multiple addresses resolving to the same user by using the format `<username>+<tag>@<domainname>`, e.g. `satoshi+whitepaper@bitcoin.org` (see [Plus signs ("+") in email addresses](https://people.cs.rutgers.edu/~watrous/plus-signs-in-email-addresses.html)). The normal flow does not change: Upon seeing such an address, `WALLET` makes a `GET` request to `https://<domain>/.well-known/lnurlp/<username>+<tag>` endpoint.

It is upon the `SERVICE` to decide what to do. If it supports tags, it SHOULD strip the `+<tag>` part and continue as normal.
A `SERVICE` MAY subsequently include the tag in the `metadata` property of the response:
```Typescript
[
        "text/tag", // optional
        string // the tag that has been passed in `<username>+<tag>`
],
```  

Currently, the `<username>` character set is limited to `a-z0-9-_.` which means the `+` sign has not been allowed yet and wallets that are not aware of it will either just deny it, or support this change and immediately use `<username>+<tag>` as `<username>`. I suppose that most wallets do not validate for `a-z0-9-_.` and will happily use everything up to `@` as username. The only tested wallet is [Zeus](https://github.com/ZeusLN/zeus), which supports sending usernames with `+`.

Example of a service that already accept `+tag`: https://ln.tips/.well-known/lnurlp/fiatjaf+test  (albeit without resolving to "fiatjaf" and a dedicated entry in metadata)
Example of a service that does not support `+tag`: https://stacker.news/.well-known/lnurlp/fiatjaf+test

With all that said, I propose adding `+<tag>` as an OPTIONAL feature with a suggestion for `SERVICE`s to handle them accordingly. No already existing `SERVICE` is forced to update and can continue to work as is.

### Benefits
With this suggestion, a `SERVICE` can offers users a "standardized" way to use multiple identifiers. e.g. a user can include `fiatjaf+s01e01@ln.tips` in the description of the first episode of a podcast, `fiatjaf+s01e02@ln.tips` for the second and so on and so forth. For written text it might be `fiatjaf+article1@ln.tips`, for code repositories it might be `fiatjaf+poncho@github.com` or `fiatjaf+nos2x@github.com`, etc.

### Drawbacks
Since a user knows whether a `SERVICE` supports this scheme, an identifier with `+<tag>` is used only when this is the case - so there is no disadvantage on the `SERVICE` side. On the `WALLET` side, it might be that the GET request will not be performed if the wallet follows the current validation rules. In this case, a user SHOULD strip the tag manually - which is arguably not desirable.

---

Should this be an addition to LUD-16 or would you rather suggest this being a distinct LUD? Imho, an own LUD seemed overkill as the changes would be minimal and optional. What do you think? 
Any and all suggestions are highly appreciated :pray:

theborakompanioni:
Resolves #197.

Adds support for `+<tag>` when paying to static internet identifiers.

### DIFF
--- a/16.md
+++ b/16.md
@@ -44,6 +44,6 @@ A `SERVICE` SHOULD strip the `+<tag>` part and continue as normal and MAY add a 
 ```
 [
     "text/tag", // optional
-    string // the tag that has been passed in `<username>+<tag>`
+    content // the `<tag>` that has been passed in `<username>+<tag>`
 ]
 ```
--- a/16.md
+++ b/16.md
@@ -7,7 +7,7 @@ LUD-16: Paying to static internet identifiers.
 
 ## Paying to [internet identifiers](https://datatracker.ietf.org/doc/html/rfc5322#section-3.4.1) (email-like addresses)
 
-The idea here is that a `SERVICE` can offer human-readable addresses for users or specific internal endpoints that use the format `<username>@<domainname>`, e.g. satoshi@bitcoin.org. A user can then type these on a `WALLET`. The `<username>` is limited to `a-z0-9-_.`. Please note that this is way more strict than common email addresses as it allows fewer symbols and only lowercase characters.
+The idea here is that a `SERVICE` can offer human-readable addresses for users or specific internal endpoints that use the format `<username>@<domainname>`, e.g. satoshi@bitcoin.org. A user can then type these on a `WALLET`. The `<username>` is limited to `a-z0-9-_.` (and `+` sign if it supports tags). Please note that this is way more strict than common email addresses as it allows fewer symbols and only lowercase characters.
 
 Upon seeing such an address, `WALLET` makes a GET request to `https://<domain>/.well-known/lnurlp/<username>` endpoint if `domain` is clearnet or `http://<domain>/.well-known/lnurlp/<username>` if `domain` is onion. For example, if the address is `satoshi@bitcoin.org`, the request is to be made to `https://bitcoin.org/.well-known/lnurlp/satoshi`.
 
@@ -32,3 +32,18 @@ or
 ```
 
 The `text/email` entry MUST be used if the internet identifier corresponds to an actual email address.
+
+### Tags
+
+A `SERVICE` can offer multiple identifiers resolving to the same user by using the format `<username>+<tag>@<domainname>`.
+
+Upon seeing such an address, `WALLET` makes a GET request to `https://<domain>/.well-known/lnurlp/<username>+<tag>` endpoint.
+
+A `SERVICE` SHOULD strip the `+<tag>` part and continue as normal and MAY add a `text/tag` entry to the `metadata` JSON array:
+
+```
+[
+    "text/tag", // optional
+    string // the tag that has been passed in `<username>+<tag>`
+]
+```
--- a/16.md
+++ b/16.md
@@ -7,7 +7,7 @@ LUD-16: Paying to static internet identifiers.
 
 ## Paying to [internet identifiers](https://datatracker.ietf.org/doc/html/rfc5322#section-3.4.1) (email-like addresses)
 
-The idea here is that a `SERVICE` can offer human-readable addresses for users or specific internal endpoints that use the format `<username>@<domainname>`, e.g. satoshi@bitcoin.org. A user can then type these on a `WALLET`. The `<username>` is limited to `a-z0-9-_.` (and `+` sign if it supports tags). Please note that this is way more strict than common email addresses as it allows fewer symbols and only lowercase characters.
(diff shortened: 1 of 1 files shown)


dni (approved):
nice addition

jklein24:
+1 - we actually added `+` to the UMA address spec for this exact use case :-)

hsjoberg (approved):
I like this.
Although this is a breaking change as current LUD-16 implementers would not accept Lightning Addresses with a tag.

theborakompanioni:
> I like this. Although this is a breaking change as current LUD-16 implementers would not accept Lightning Addresses with a tag.

You are perfectly right, however, as described in #197:

> Since a user knows whether a `SERVICE` supports this scheme, an identifier with `+<tag>` is used only when this is the case - so there is no disadvantage on the `SERVICE` side. 

So in practice this should not lead to any problems on the `SERVICE` side (as users would just not use the feature).

It is a different story on the `WALLET` side, if it strictly implements previous validation:
> On the `WALLET` side, it might be that the `GET` request will not be performed if the wallet follows the current validation rules. In this case, a user SHOULD strip the tag manually - which is arguably not desirable.

In practice, some wallets do not validate the address and just perform the `GET` request as is, e.g. tested with wallet is [Zeus](https://github.com/ZeusLN/zeus), which supported sending usernames with `+` (at the time of writing).

What do you think?

dni:
> I like this. Although this is a breaking change as current LUD-16 implementers would not accept Lightning Addresses with a tag.

i think current implementation would just see it as new address and would probably just throw a 404 or similar, but that is on the user to not use the tag on an unsupporting service. its also simple for the user to test if a service supports it, just try fetching an invoice with that address + tag.

andrerfneves:
Ack. This is good 👍

hsjoberg:
> It is a different story on the WALLET side

@theborakompanioni Yes, I was mostly concerned about `WALLET`s implementing it strictly.

Anyway I don't think this is a big deal.

I'm for merging this one.

16.md
```diff
@@ -32,3 +32,18 @@ or
...
+A `SERVICE` can offer multiple identifiers resolving to the same user by using the format `<username>+<tag>@<domainname>`.
+
+Upon seeing such an address, `WALLET` makes a GET request to `https://<domain>/.well-known/lnurlp/<username>+<tag>` endpoint.
+
+A `SERVICE` SHOULD strip the `+<tag>` part and continue as normal and MAY add a `text/tag` entry to the `metadata` JSON array:
+
+```
+[
+    "text/tag", // optional
+    string // the tag that has been passed in `<username>+<tag>`
```
hsjoberg:
```suggestion
    string // the `<tag>` that has been passed in `<username>+<tag>`
```

nit: trying to avoid confusion on what this metadata field should include. If you quickly read it you might think it should be `<username>+<tag>`

dni: small nitpick,  "text/email" above has `content` instead of `string` maybe for consistency change it to content.

16.md
```diff
@@ -7,7 +7,7 @@ LUD-16: Paying to static internet identifiers.
 
 ## Paying to [internet identifiers](https://datatracker.ietf.org/doc/html/rfc5322#section-3.4.1) (email-like addresses)
 
-The idea here is that a `SERVICE` can offer human-readable addresses for users or specific internal endpoints that use the format `<username>@<domainname>`, e.g. satoshi@bitcoin.org. A user can then type these on a `WALLET`. The `<username>` is limited to `a-z0-9-_.`. Please note that this is way more strict than common email addresses as it allows fewer symbols and only lowercase characters.
+The idea here is that a `SERVICE` can offer human-readable addresses for users or specific internal endpoints that use the format `<username>@<domainname>`, e.g. satoshi@bitcoin.org. A user can then type these on a `WALLET`. The `<username>` is limited to `a-z0-9-_.` (and `+` sign if it supports tags). Please note that this is way more strict than common email addresses as it allows fewer symbols and only lowercase characters.
```
dni:
```suggestion
The idea here is that a `SERVICE` can offer human-readable addresses for users or specific internal endpoints that use the format `<username>@<domainname>`, e.g. satoshi@bitcoin.org. A user can then type these on a `WALLET`. The `<username>` is limited to `a-z0-9-_.` (and `+` if the `SERVICE` supports tags). Please note that this is way more strict than common email addresses as it allows fewer symbols and only lowercase characters.
```

16.md
theborakompanioni: Good catch! Changed!  Thanks! :pray:

16.md
theborakompanioni: Changed! :muscle:

dracos4lyfe:
How do I start creating my own codes

theborakompanioni:
> How do I start creating my own codes

The idea is that you should be able to append whatever you like to your existing lightning address. They do not really exist and are "created on the fly" when people use them.

dni (approved):

andrerfneves:
Implementing this in ZBD side asap

PR #280, opened 2025-07-28
Add Piggy service and protocol support references

realbub:
Updated README to include Piggy as a supported service and referenced its support for multiple protocol specs, including payRequest, static internet identifiers, payer identity, and long payment descriptions. Added Piggy's link to the list of wallet and service providers.


hsjoberg:
Great! 💯

PR #281, opened 2025-08-19
Add long-held HTTP request support to LUD21 verify

elsirion:
Polling is an antipattern in most cases, it increases latency for clients and server load, yet it's the only way to use LNURL verify aka LUD21 right now. When e.g. building a PoS using LUD21 one has to poll at least every second to achieve reasonably low latency.

Long-polling fixes this.

### DIFF
--- a/21.md
+++ b/21.md
@@ -53,3 +53,8 @@ or
   "reason": "Not found"
 }
 ```
+
+## Optional Long-Held HTTP Mode
+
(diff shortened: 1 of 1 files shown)


Kukks (changes requested):
Longpolling is kind of stale tech by now. 

SSE seems to be the goto approach today.

I'd also perhaps expect some kind of signalling capability to know we can do this?

andrerfneves:
Agree on the SSE being the more stable modern path forward, but open to suggestions here. I do think solving over-polling is great and something we should aim to achieve, but the spec additions as they are, feel very bare bones.

dni (approved):

PR #285, opened 2025-11-18
Adds Cash App supporting LUD 06 and LUD 16

hugomd:
Cash App supports sending/receiving via LUD 16, and paying LNURLs (`lightning:lnurl...`).


rgrnwd (approved):

hsjoberg:
Great! 💯

realbub:
I believe there’s a bug with the minimum amounts, possibly due to an incorrect conversion from mSats to sats. For example, at https://pig.gy/ we have a minimum of 100 sats, but Cash App doesn’t allow sending less than 10,000 sats. You can test it here: https://pig.gy/saintosh.i

hugomd:
> I believe there’s a bug with the minimum amounts, possibly due to an incorrect conversion from mSats to sats. For example, at [pig.gy](https://pig.gy/) we have a minimum of 100 sats, but Cash App doesn’t allow sending less than 10,000 sats. You can test it here: [pig.gy/saintosh.i](https://pig.gy/saintosh.i)

I've just had a look at this, Cash App is respecting the minimum amount of 100 sats, and allows me to initiate for that amount, under 10,000 sats.

How are you attempting the payment from Cash App?

PR #287, opened 2026-04-09
Add Lexe supporting LUD-06, LUD-12, and LUD-16

nicolft:
[Lexe](https://www.lexe.app/) supports LUD-06, LUD-12, and LUD-16!


hsjoberg:
Great! 💯

PR #294, opened 2026-05-30
Improve LUD-04: add linkingKey derivation summary and early references

### LINKED ISSUE #262: LUD-04 seems incomplete
Taking a look at https://github.com/lnurl/luds/blob/luds/04.md , there is no detailed explanation for the `linkingKey` derivation. I see there is an older document https://github.com/lnurl/luds/blob/legacy/lnurl-auth.md which contains a lot more detail.

I look harder and I find LUD-05 and LUD-13, but those documents are never mentioned in LUD-04. I think there should be some clearer ties between them. Also, seems like the graphic from the legacy document is gone. Are there any other missing LUD references that I should be looking at to more clearly understand LNURL-auth?

andrerfneves:
Fixes #262

## Changes

- **Add early reference to LUD-05 and LUD-13** in the intro paragraph so readers immediately know where to find the derivation details
- **Add a BIP32-based derivation summary** — the three-step process (m/138'/0 hashingKey → HMAC-SHA256 with domain → m/138'/<long1>/<long2>/<long3>/<long4>) that was present in the legacy doc but missing from the current LUD
- **Add signMessage-based derivation summary** with a link to LUD-13 for wallets that cannot access the master private key
- **Fix JSON code block language tags** to lowercase (json) for consistency with other LUDs

## Why

The legacy lnurl-auth.md had detailed derivation logic (including a Scala code example) that got lost when the spec was split into LUD-04, LUD-05, and LUD-13. The current LUD-04 just says "go read these other docs" without giving any conceptual context. This adds back the high-level derivation overview so readers can understand the flow before diving into the full specs.

### DIFF
--- a/04.md
+++ b/04.md
@@ -9,6 +9,8 @@ LUD-04: `auth` base spec.
 
 A special `linkingKey` can be used to login user to a service or authorise sensitive actions. This preferably should be done without compromising user identity, so plain LN node key can not be used here. Instead of asking for user credentials, a service could display a "login" QR code which contains a specialized `LNURL`.
 
+The `linkingKey` is a domain-specific key derived from the user's wallet seed. For details on how it is derived, see [LUD-05](05.md) (BIP32-based) and [LUD-13](13.md) (signMessage-based).
+
 
 ### Server-side generation of auth URL and signature verification:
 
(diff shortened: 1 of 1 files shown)


andrerfneves:
@Kukks small doc improvement on LUD04 explanation linking to 05 and 13. If all good with you let's merge.

Kukks (approved):

PR #296, opened 2026-06-01
LUD-16: add optional default identifier shorthand

andrerfneves:
## Motivation

Lightning Address works well for hosted services, but personal domains often end up using placeholder local parts like `me@domain`, `pay@domain`, or `ln@domain` even when the domain itself is the intended identifier. This came up in andrerfneves/lightning-address#28.

## Proposal

This extends LUD-16 with an entirely optional default identifier convention:

- services MAY expose `_@domain` as the canonical default identifier
- wallets MAY accept `@domain` as shorthand for `_@domain`
- resolution uses the existing LUD-16 endpoint shape: `/.well-known/lnurlp/_`
- wallets and services that do not implement the shorthand remain fully compatible

This avoids introducing a new endpoint shape and keeps metadata aligned with the existing LUD-16 `user@site.com` identifier format.

## Testing

Docs-only spec change; no tests run.

### DIFF
--- a/16.md
+++ b/16.md
@@ -13,6 +13,16 @@ Upon seeing such an address, `WALLET` makes a GET request to `https://<domain>/.
 
 The response from `SERVICE` then MUST be the same as in [LUD-06](06.md), step 3, and the flow is the same.
 
+### Default identifier
+
+A `SERVICE` MAY offer a default internet identifier for a domain by using `_` as the `<username>`.
+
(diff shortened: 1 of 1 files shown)


Kukks (approved):

dni (approved):

hsjoberg (approved):

dni:
https://github.com/lnbits/lnurl/commit/4dd00d5f062ed74d739529348d4f96a7b2873f80

PR #297, opened 2026-06-06
LUD-XX: nodePubkeys for payRequest

andrerfneves:
Adds an optional `nodePubkeys` field to the LUD-06 payRequest response, allowing services to advertise the Lightning node public keys that may generate invoices for their LNURL endpoint.

**What it does:**
- A wallet fetches the LNURL, gets back `nodePubkeys` alongside the existing fields
- When the wallet receives the invoice from the callback, it extracts the payee node pubkey (from the BOLT11 `n` field or signature recovery)
- If the invoice's payee doesn't match any advertised pubkey, the wallet shows a non-blocking warning
- If `nodePubkeys` is absent, behavior is unchanged from LUD-06

**Why this matters:**
Invoices are increasingly presented through abstractions (embedded browsers, platform relays, aggregated UIs) where the TLS connection is hidden from the user. Wallets also show less invoice detail — often just an amount. This lets the wallet verify invoice provenance at a layer the user can reason about: "this invoice came from one of the nodes this service owns."

**Design:**
- `nodePubkeys` is an array of objects (not raw strings) — each entry has a mandatory `pubkey` field, with room for future fields like `alias`, `since`, `signature`, `features` defined by later LUDs
- Uses `LUD-XX` as a placeholder — needs numbering resolution before merge

**Dependency:** `06 -> XX`


dni:
note: in paylink the pubkey could be recovered from the bolt11

PR #298, opened 2026-06-18
LUD-XX: Currencies

ethanrose:
My initial PR (Pouch.ph implementation) was https://github.com/lnurl/luds/pull/207

This new proposal is an update from my original PR, in order to be compatible with:

- Bipa implementation: https://github.com/lnurl/luds/pull/251

- LightSpark implementation: https://github.com/uma-universal-money-address/protocol/blob/main/umad-04-lnurlp-response.md

However, I have left out some of their implementation details in the spirit of keeping the spec as barebones as possible. 

Namely:

- **Convert/convertible**: At Pouch, currency conversion is a recipient config, not a sender's payment flow decision. If some services wish to give the sender power over the recipient's settlement currency by sending a param, it should be a follow-on LUD.

- **Final Response Multiplier & Fee** This doesn't provide info that the wallet can't already derive (multiplier) or should trust (fee). If services are using this, it should be a follow-on LUD.

### DIFF
--- a/24.md
+++ b/24.md
@@ -76,4 +76,4 @@ If there is a `currency` record, `WALLET` should display a modified interface:
 
 ## 5. User interface for reviewing & confirming the exchange rate
 
-- `WALLET` must display a confirmation screen so that the user can review and confirm the invoice amount.
+`WALLET` must display a confirmation screen so that the user can review and confirm the invoice amount.
--- a/24.md
+++ b/24.md
@@ -17,7 +17,7 @@ This document describes an extension to the [payRequest](https://github.com/lnur
 
 The main features provided by this extension are:
 - `SERVICE` **MAY** provide `WALLET` with a list of supported currencies.
-- `WALLET` **MAY** allow user to pay an amount denominated in a supported currency.
+- `WALLET` **MAY** allow user to request an invoice for an amount denominated in a supported currency in place of millisats.
 
 
 ### 1. `currencies` array in payRequest details
--- a/24.md
+++ b/24.md
@@ -1,0 +1,79 @@
+LUD-24: `currencies` base spec.
+=======================================
+
+`author: ethanrose` `author: lsunsi` `author: lorenzolfm` `author: jklein24`
+
+---
+
+## Support for LNURL-pay currencies
+
+This document describes an extension to the [payRequest](https://github.com/lnurl/luds/blob/luds/06.md) base specification that allows the `WALLET` to send money to a `SERVICE` while denominating the amount in a different currency. The features proposed enable many use cases including:
+
+- Bitcoin-settled cash remittances
+- Merchant bitcoin acceptance
+- Lightning address aliases for bank & mobile money accounts
+- Cross-border payroll
+
+
+The main features provided by this extension are:
+- `SERVICE` **MUST** inform `WALLET` what currencies it supports
+- `WALLET` **MAY** request an invoice with an amount denominated in one of the currencies
+
+
+## 1. `currencies` array in payRequest details
+
+`SERVICE` must alter its JSON response to the first request to include a `currencies` field, as follows:
+
+```diff
+{
+  "tag": "payRequest",
+  "callback": string,
+  "maxSendable": number,
+  "minSendable": number,
+  "metadata": string,
++ "currencies": [
++   {
++     "code": "PHP",
++     "name": "Philippine Pesos",
++     "symbol": "₱",
++     "decimals": 2,
++     "minSendable": 1,
++     "maxSendable": 50000,
++     "multiplier": 64501 // (optional)
++   }
++ ]
+}
+```
+
+- The inclusion of the `currencies` field indicates the support of this extension
+- The `multiplier` may be included for informational purposes, not a guaranteed rate
+- The `code` of a `currency` must be unique
+- The order of the `currencies` may be interpreted by the `WALLET` as the receiving user preference for a currency
+
+
+## 2. Wallet interface
+
+If there is a `currency` record, `WALLET` should display a modified interface:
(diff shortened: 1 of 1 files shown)


ethanrose:
For review, if everyone could give either a 👍 or express any blocking concerns: @callebtc @trbouma @lsunsi @fiatjaf @vindard @jklein24 @lorenzolfm @kukks @kngako (please feel free to mention others for review as well -- several others are using a variant of this spec, but I am unsure of their github handles).

24.md
```diff
@@ -0,0 +1,81 @@
...
++     "minSendable": 1,
++     "maxSendable": 50000,
++     "multiplier": 64501 // (optional)
++   }
++ ]
+}
+```
+
+- The inclusion of the `currencies` field indicates the support of this extension
+- The `multiplier` may be included for informational purposes, not a guaranteed rate
```
jklein24, line 49: This is the reason that the multiplier in the payreq response is critical IMO. The way this is interpreted in UMA is that the first response includes an estimate for while the sender is typing an amount, then the second response actually locks a quote including rate and fees.

24.md
```diff
@@ -0,0 +1,81 @@
...
+- Display currency symbol and code
+- If there are multiple currency options, allow user to select/change currencies
+- Validate "decimals" count for user amount input
+
+
+### 3. Including `currency` in callback parameters
+
+```diff
+- <callback><?|&>amount=<number>
++ <callback><?|&>currency=PHP&amount=<numberInSmallestUnit>
```
jklein24, line 66: This is missing critical functionality of the original LUD which allows the sender to lock either the sending *or receiving* amount. If I'm sending from USD to MXN, I need to be able to say either "I'm sending $100 USD" or "I need the receiver to receive $100 MXN". Both are quite important for different use cases. UMA doesn't use the full subset of flexibility for the `convert` stuff, but it definitely needs this point to still exist.

jklein24 (commented):
I don't think I agree that the stuff you're leaving out should be follow-up LUDs. They're pretty critical functionality for us at least.

lsunsi:
I haven't had the time to read the full spec yet, but I'm inclined to agree with @jklein24 comment's about the `quote` mechanism being core to the functionality.

The current implementation at [showcased here](https://lnurlp-live.vercel.app/lsunsi@bipa.app) shows that working, if you "receive as Reais R$" you'll be able to see rich information about a quote (price, fee and how much sats it'd cost) with a ever-refreshing invoice but fixed receiving amount in local currency.

24.md
```diff
@@ -0,0 +1,81 @@
...
+
+`author: ethanrose` `author: lsunsi` `author: lorenzolfm` `author: jklein24`
+
+---
+
+## Support for LNURL-pay currencies
+
+This document describes an extension to the [payRequest](https://github.com/lnurl/luds/blob/luds/06.md) base specification that allows the `WALLET` to send money to a `SERVICE` while denominating the amount in a different currency. The features proposed enable many use cases including:
+
+- Bitcoin-settled cash remittances
```
lorenzolfm, line 12: What do you understand by "remittances"?

From what I understood from the spec, this doesn't enable a remittance UX like Wise, for instance. This would required the convertible info

24.md
```diff
@@ -0,0 +1,81 @@
...
+  "metadata": string,
++ "currencies": [
++   {
++     "code": "PHP",
++     "name": "Philippine Pesos",
++     "symbol": "₱",
++     "decimals": 2,
++     "minSendable": 1,
++     "maxSendable": 50000,
++     "multiplier": 64501 // (optional)
```
lorenzolfm, line 42: Why do think this should be optional?

The `service` already knows the multiplier because its needed in order to fulfill the denomination request. So why not share it with the `wallet`.

24.md
lorenzolfm, line 66: Why did you go with currency outside instead of inside the amount, like [here](https://github.com/bipa-app/luds/blob/2446ff5a9ed51fcec8079f4a3e8a9539df857231/22.md?plain=1#L98)

Is not that we don't like this option, but we chose the other one because without it was less umbiguous in case `wallet` misuses the API. A service not supporting the LUD wouldn't even be able to understand the request

24.md
```diff
@@ -0,0 +1,81 @@
...
+  "minSendable": number,
+  "metadata": string,
++ "currencies": [
++   {
++     "code": "PHP",
++     "name": "Philippine Pesos",
++     "symbol": "₱",
++     "decimals": 2,
++     "minSendable": 1,
++     "maxSendable": 50000,
```
lorenzolfm, line 41: Given that multiplier is not optional (see [here](https://github.com/lnurl/luds/pull/298/changes#r3499218875)), wouldn't `minSendable` and `maxSendable` be a derivation of `multiplier` and top level `minSendable` and `maxSendable`.

Since this LUD is just about *denomitation* there is no further limitation imposed by different currencies not already represented in the `minSendable` and `maxSendable` on top level

lorenzolfm (commented):
Hey @ethanrose. Just pair reviewed the PR with @lsunsi.

lsunsi:
Just confirming that I concur with @lorenzolfm 's questions, and I have suggestion.

Although I think #251  was good to go, I think slicing #251 with just the denomination functionality might be useful in order to get it merged. That said, I'd be more comfortable endorsing this approach if it was compatible with #251 so it could be PRd later with the convertibility functionality.

We are quite close but not quite there and answering @lorenzolfm 's questions on the subject will get us closer.

- multiplier being mandatory
- min/max inside currency being removed (implied by multiplier and min/max from millisats)
- maybe changing the request format (I think this doesn't hurt compatibility, so it's not necessary)

PR #300, opened 2026-08-04
LUD-XX: Ark address in payRequest response

Codepocketdev:
Adds an optional `ark` field to the LUD-06 payRequest response, enabling wallets to route payments via arkoor (zero-fee, instant) when sender and recipient are on the same Ark server, falling back to Lightning otherwise.

## Summary

- Adds optional `ark` field (bech32 Ark address) to payRequest response
- Adds optional `ark_server_pubkey` field for explicit server verification
- Fully backwards compatible — wallets that don't implement this ignore the fields
- Live demo: `martin@pay.satoshipay.me` returns both fields
- Noah wallet already reads the `ark` field and routes via arkoor (confirmed mainnet test, 0 fee)

## Live test

```
curl https://pay.satoshipay.me/.well-known/lnurlp/martin
```

Returns ark and ark_server_pubkey fields. Noah wallet paid 5 sat to this address at 0 fee via arkoor on Bitcoin mainnet.

## Reference implementation

Arkoor routing is implemented in cashubtc/cdk-payment-processors (bark crate).

## Discussion

Developed in collaboration with engineers from Second (the Ark protocol team).

### DIFF
--- a/XX.md
+++ b/XX.md
@@ -1,0 +1,85 @@
+LUD-XX: Ark address in payRequest response.
+=============================================
+
+author: Codepocketdev | discussion: TBD
+
+---
+
+## Motivation
+
+Wallets that support the Ark protocol can route payments via
+arkoor (out-of-round) transactions, achieving zero-fee instant
+settlement between users on the same Ark server. Currently,
+there is no standard way for an LNURL-pay service to advertise
+an Ark address alongside a Lightning address, forcing all
+payments through the Lightning gateway even when a zero-fee
+Ark-native path is available.
+
+## Specification
+
+LN SERVICE MAY include two optional fields in the payRequest
+response defined in LUD-06:
+
+  callback: string
+  maxSendable: number
+  minSendable: number
+  metadata: string
+  tag: payRequest
+  ark: string               (optional: bech32 Ark address)
+  ark_server_pubkey: string (optional: hex-encoded Ark server pubkey)
+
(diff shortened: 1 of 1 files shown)


hsjoberg:
I think we should not short-circuit the LNURL-pay protocol by providing the ark address in the first `SERVICE` response.
Rather, I propose that we add it in the second response, so that protocols such as [LUD-12](https://github.com/lnurl/luds/blob/luds/12.md) comments, [LUD-18](https://github.com/lnurl/luds/blob/luds/18.md) payerid and other protocol extensions work. This is getting implemented in Noah in https://github.com/smolcars/noah/pull/309.

https://github.com/hsjoberg/lads/blob/master/01.md

Codepocketdev:
Closing in favor of LADS-01 (https://github.com/hsjoberg/lads/blob/master/01.md) which handles this more correctly with a two-step negotiation. Will implement LADS-01 instead.

hsjoberg:
Hi @Codepocketdev, there's ongoing work to make a protocol extension that isn't Ark-specific. #303

So I think we shouldn't pursue neither my proposal nor yours. Instead let's see when the dust settles on #303 and we can make sure Noah, Satoshipay and others supporting Second's bark ASP stay compatible with each other.

Codepocketdev:
Yeah  I had closed my PR ,I  guess will follow up with PR #303 and find out
whats next.

PR #301, opened 2026-08-06
LUD-25: `LNURLcash`, Bearer assets

dni:
LNURL-withdraw ([LUD-03](03.md)) links are redeemable by anyone who holds a valid `k1`. This document formalizes that property into a bearer-asset scheme: a `k1` is no longer only a one-time ticket that gets exchanged for a BOLT-11 invoice, it can *itself* be the asset. A `k1` that a `SERVICE` has credited with value can be copied, printed, NFC-tapped or handed over exactly like a banknote, and can be re-issued ("melted and re-minted"), merged with other bearer notes, or split into change, all without the holder needing to be online or to have a route back to `SERVICE` at handoff time.

Because a bearer `k1` is nothing more than a normal `withdrawRequest` `k1`, `LNURLcash` requires **no new endpoint and no new encoding**. A `SERVICE` that implements it is still speaking plain [LUD-03](03.md); a `WALLET` that does not know about `LNURLcash` still sees a normal withdraw link and can cash it out to a BOLT-11 invoice as usual. `LNURLcash` is therefore fully optional and backward-compatible.

## projects
i am currently working on a lnurl-wallet and lnurl-mint. passed alpha now in beta testing

- https://github.com/dni/lnurl-wallet / https://wallet.lnurlcash.com
- https://github.com/dni/lnurl-mint / https://mint.lnurlcash.com
- https://github.com/dni/lnurl-vault / https://vault.lnurlcash.com


dni:
hey wow very cool, i though about most of this myself. expiry would be cool and there should be a hardcap on urls. thank you very valid review.

TheCryptoDonkey:
One gap: nothing in the draft says a burn is permanent.

The outstanding-set paragraph at the end of *Redeeming a bearer note* says what changes which `k1`s are outstanding, but nothing forbids `SERVICE` from later crediting a burned `k1` back into that set. The lifecycle is unidirectional by construction rather than by rule, so a mint that stores notes as a mutable balance keyed by `k1` can refill a spent one and still be compliant.

Worth stating explicitly because the wrong version is genuinely attractive: print notes on paper, refill them on a timer, claimable forever. That reopens the previous-holder window *Offline circulation* already warns about, permanently, and for every holder the note has ever had rather than just the last one.

Suggested, appended to that paragraph:

> A burn is permanent: once a `k1` is burned, `SERVICE` MUST NOT return it to the outstanding set, credit it with new value, or reissue it. The only exception is restoring a *pending* `k1` whose melt failed to settle, which never left the set.

TheCryptoDonkey:
Another gap, and this one the draft actively mandates.

A rotate/split/merge is a GET. HTTP treats GET as idempotent, so stacks retry it on a dropped connection. An `LNURLcash` mutation is not idempotent: the first attempt already burned the input.

The draft then requires the retry to be answered:

> As with any `withdrawRequest`, an invalid, already-spent, or unknown `k1` MUST yield `{"status": "ERROR", "reason": string}`

A `WALLET` reads that as definitive: the request was refused, so the fresh preimage behind `h` minted nothing and can be dropped. But `SERVICE` did mint it, on the first attempt, and that preimage was the only copy in existence. The note is gone, and the `WALLET` was never told anything happened to its money, only that its `k1` was stale.

This isn't hypothetical, and it isn't something you can just tell implementers to avoid:

- Go's `net/http` retries a GET that failed on a reused idle connection. Avoiding it needs `DisableKeepAlives: true` on a transport of your own, since the process-wide `DefaultTransport` pool will otherwise supply exactly that condition.
- Java's `java.net.http.HttpClient` retries idempotent methods on a dropped connection and exposes no switch to stop it (`jdk.httpclient.disableRetryConnect` governs connect failures only; setting it changes nothing here). A JVM wallet has to abandon the JDK client outright. I ended up on OkHttp with `retryOnConnectionFailure(false)`.

Both surfaced by hanging up mid-mutation against a mock mint, and both turned an ambiguous outcome into a confident, wrong "already spent".

**The fix is cheap, because the `WALLET` supplies `h`.** A replay is fully self-identifying: same `k1` set, same `h`/`h2`, same `amount`. `SERVICE` can tell it apart from a double-spend and answer with the original success.

Suggested, after that ERROR sentence:

> Because the `WALLET` supplies `h`, a repeated request is distinguishable from a double-spend attempt. If a rotate, split or merge names `k1`(s) already burned by an earlier request carrying the identical `h` (and `h2`, and `amount`), `SERVICE` SHOULD return that earlier request's `{"status": "OK"}`, including the same `sig`/`sig2`, rather than an error. Transports retry GETs, and without this a retried mutation is reported to the `WALLET` as a definitive failure, causing it to discard the only copy of a note `SERVICE` has already minted. A request naming burned `k1`(s) with any *other* `h` is a genuine double-spend attempt and MUST still fail.

Replaying discloses nothing: `h` is a hash the requester chose and already holds, `sig` is over `h`, and the preimage never leaves the `WALLET`. Someone replaying a captured URL learns only that it once succeeded, and still can't spend the output.

Cost to `SERVICE` is one association per burn, the `k1`s burned to the `h` they minted, recorded on rows it already keeps for spent notes.

Melt is less urgent, since a repeat hits the `pending` rule and there's no `WALLET`-generated secret to lose, but the same rule would make that retry harmless too.

TheCryptoDonkey:
We've built an independent LUD-25 stack against this draft — a mint, a wallet library, and a language-neutral conformance suite with an adversarial mock mint — and a security pass surfaced a handful of things the draft currently leaves unsaid. Each is something we either got wrong ourselves before catching it, or watched a perfectly ordinary HTTP stack get wrong on our behalf. None needs new protocol machinery; they all just need saying out loud.

First, a small doc bug: the minting section says "see Security considerations" twice, but the PR doesn't add one. The notes below are our offer of what it could contain.

**1. Mutations are non-idempotent GETs — wallets must not retry, caches must not replay**

Every mutation is an HTTP GET, and HTTP treats GET as safe and retryable. Stacks act on that assumption on their own initiative: browsers resend GETs on stale pooled connections, service workers and retry interceptors re-issue them, proxies and meshes auto-retry "idempotent" requests, and one popular DOM-test library physically sends every GET twice. The first attempt burns the input `k1`; the retry is answered "already spent", and a wallet that believes that answer discards the fresh secret that is the only copy of the note the first attempt minted. This broke two of our own implementations during development, by two different mechanisms.

Suggested normative text:

> A `WALLET` MUST NOT automatically retry a callback request. On any uncertain outcome (timeout, dropped connection, unparseable or absent confirmation) it probes the informational GET to learn the truth before concluding anything. A `SERVICE` SHOULD refuse any non-GET request to its endpoints — an `OPTIONS` preflight carrying a callback's query string must never mutate anything — and SHOULD answer callback requests with `Cache-Control: no-store`, so a shared cache cannot replay a mutation response the `SERVICE` never saw.

**2. Redirects on k1-bearing URLs**

None of LUD-01/03/25 say whether a wallet follows a 3xx. For most LNURL flows that's a robustness question; for bearer notes it's a secret leak, because the callback URL carries `k1`. An https endpoint can legitimately redirect, and nothing stops the target being `http://` cleartext or another host — the wallet's HTTP layer will happily follow, sending the bearer secret in the clear.

Suggested text:

> A `WALLET` following a redirect MUST re-admit every hop against the same rules as the original URL (https, or http only to loopback/`.onion`), or refuse to follow. A `SERVICE` SHOULD NOT answer LNURL requests with redirects at all.

**3. A callback refusal proves nothing about a k1**

The draft requires `ERROR` for "invalid, already-spent, or unknown" k1, but the refusal is atomic and necessarily ambiguous: the same response also covers a duplicated `k1`, a malformed one, and an output hash (`h`/`h2`) that collides with an existing note. A wallet that locks a note as spent on the strength of a callback refusal can brick a note that is perfectly live.

Suggested text:

> The informational GET is the only authoritative probe of a note's state. A `WALLET` MUST NOT conclude from a callback refusal that a `k1` is spent; it confirms spent-vs-live with a fresh informational GET.

**4. Sharing a funding source between mints**

Nothing in the draft considers two mints backed by one Lightning node, but it's an attractive deployment (one node, several small mints) and it opens a replay no mint's own tables can see: an invoice melted at mint A is unknown to mint B, whose payment dedupe-by-hash reads A's completed payment as its own success and burns B's note with no funds moving.

Suggested text:

> A `SERVICE` whose funding source may be shared MUST refuse a melt whose payment hash the node has already paid or is still paying — not merely one its own records know — and MUST refuse to pay an invoice it issued itself.

**5. A deployment note on fees**

The draft says the mint fee covers the routing cost of the eventual payout. Worth one sentence for operators: a fee-free `SERVICE` still pays routing on every melt, and minting-and-melting at minimum amounts is costless to a griefer, so operators should charge a base fee that at least covers their routing budget.

Happy to turn any of these into suggested prose directly on the document if that's easier to review.

TheCryptoDonkey:
**6. Which spelling is `withdrawLink`?**

The minting section says `withdrawLink` is "a raw, non bech32-encoded URL as described in LUD-17". LUD-17 describes two things: the `lnurlw://` scheme, and the plain `https://` URL it stands for. The diagram shows `https://mint.example/w`; the next sentence says appending `?k1=` to it "yields a bearer note", which is only literally true of `lnurlw://`. Both readings shipped: lnurl-mint emits `https://mint.example/w`, our mint emits `lnurlw://mint.example/w`, and a wallet that handles one but not the other fails against real mints today.

I'd pin the plain form. Every other URL in an LNURL body (`callback`, `verify`) is directly fetchable, so a lone `lnurlw://` field is a trap for a wallet that just fetches it; the plain form is explicit about transport for onion hosts; and `lnurlw://` is what a wallet puts on the QR, a choice it makes at display time anyway.

Suggested text:

> `withdrawLink` is the plain URL of the `SERVICE`'s informational `withdrawRequest` endpoint: `https://`, or `http://` for an onion service. A `WALLET` SHOULD also accept the `lnurlw://` form ([LUD-17](17.md)) for the same value.

Either way, the conformance vectors now carry both spellings so clients are tested against both.

TheCryptoDonkey:
Re-read the document at fc1296c against my comments above, since text has landed since I wrote them. Where each stands:

**Addressed — closing from my side:**

- **Burn permanence.** Adopted verbatim in *Redeeming a bearer note* ("A burn is permanent: once a `k1` is burned, `SERVICE` MUST NOT return it to the outstanding set…"). Nothing further needed on that one.
- **The dangling "see Security considerations" references.** The section exists now, covering the no-`comment` fallback / `verify` exposure and the log-hygiene point.

**Still open:**

1. **Retry rule for mutation GETs.** The ERROR sentence in *Redeeming a bearer note* still mandates answering a retried rotate/split/merge as a definitive failure. The proposed rule — an identical replay (same `k1`s, same `h`/`h2`, same `amount`) returns the original `{"status": "OK"}`, with the same `sig`/`sig2`, rather than an error — hasn't landed. This remains the one I'd most want pinned before merge: two mainstream HTTP stacks retry GETs on their own initiative, and without the rule a transport hiccup is reported to the wallet as "already spent", so it discards the only copy of a note `SERVICE` already minted.
2. **Security-bundle items 1–5:** wallets MUST NOT auto-retry callbacks and probe the informational GET on any uncertain outcome (plus `Cache-Control: no-store` and refusing non-GET requests); a redirect policy for `k1`-bearing URLs; a callback refusal proves nothing about a `k1` (the informational GET is the only authoritative probe); a shared funding source between mints (dedupe by payment hash at the node, refuse to pay self-issued invoices); and one operator sentence on fee-free mints still paying routing on every melt. None of these are in the text yet. (The new split/merge fee rules do cover the dust-splitting side of the griefing concern.)
3. **`withdrawLink` spelling.** Still "a raw, non bech32-encoded URL as described in LUD-17", and LUD-17 names both the plain URL and the `lnurlw://` scheme — both readings are live in real mints today. My vote stays the plain form with a SHOULD-accept for `lnurlw://`, but either pin beats the ambiguity.

Separately, a couple of items on the seed-derivation section and the `h` lookup that I'll raise directly rather than in this thread.

TheCryptoDonkey:
**Line 80 and the reference implementation have diverged, and I think the
implementation has it right.**

`dni/lnurl-mint` `b257d58` makes `comment` mandatory: a missing or malformed
one is rejected rather than falling back to a preimage-keyed note.
`commentAllowed: int = 64` is a fixed field on the payRequest model, so the
mint always advertises it and always requires it. Line 80 still asks for the
fallback, so the document and the reference now describe different mints, and
it would be good to close that gap in favour of the mint.

The draft already lays out the reasoning for it: `P` propagates hop by hop as
each node settles its outgoing HTLC, so every routing node learns it,
routinely before the payer has finished processing the payment. Security
considerations says exactly this, and answers it by advising the wallet to
rotate immediately — which is a race, and the fallback is what puts the
wallet in one. Requiring the comment removes the race rather than shortening
it.

There may be a second reason coming, though I'd flag it firmly as a direction
of travel rather than a fact on the ground: dni/lnurl-mint#29, from
@aljazceru, is a draft PR — conflicts outstanding, no review yet — adding a
Spark (Breez SDK) funding source, and it notes that a spark-routed payment
completes with no preimage. That holds for Spark-native settlement
specifically; a payment that exits to Lightning through a service provider
still has one. But Spark-native is exactly the case a Spark-funded mint is in,
and if that PR or anything like it lands, the fallback isn't merely unwise on
such a backend, it's unimplementable — there's no `P` to key the note by. Any
funding source that settles without revealing a preimage would be in the same
position.

A possible replacement for line 80's first bullet:

> * `SERVICE` MUST reject a mint payment carrying no `comment`, or one that
>   isn't a bare hex-encoded 32-byte hash, with
>   `{"status": "ERROR", "reason": string}`, before any invoice is issued. A
>   note keyed by the payment preimage is redeemable by every routing node
>   that forwarded the payment, and a funding source that settles without
>   revealing a preimage cannot key one at all.

That would also make the LUD-21 `verify` restriction in the same bullet
unnecessary: with no fallback there's no case where `verify` hands out the
note rather than proof it was paid.

One knock-on worth catching in the same change — and I had this wrong myself
at first, so it's easy to miss. The line affected is 31 rather than 12. Line
31 says a wallet that knows nothing about LNURLcash "sees a normal
`payRequest`, pays it like any other, gets back a normal BOLT-11 invoice",
and that support is "entirely additive on top of that same payment". Under a
mandatory `comment` that no longer holds — such a wallet gets an error. Line
12's own example is the withdraw side, which survives untouched: line 170
still stands, and a wallet that has never heard of LNURLcash can still redeem
a note it's handed. So the compatibility claim narrows rather than
disappears, and it might be worth saying so directly: **minting** needs an
LNURLcash-aware wallet, **redeeming** stays plain LUD-03 and works with
anything. Line 12's "fully optional and backward-compatible" would want
qualifying to match.

@dni — does that match your intent for `b257d58`? If it's useful I'm happy to
open a PR against the branch with the above, or leave it to you. Thanks for
keeping the spec and the reference mint moving together — having both to read
is what made the divergence easy to spot.

dni:
@TheCryptoDonkey updated to lud and addresses most of your issues.

bitkarrot:
LFG!!!!

dni:
added @TheCryptoDonkey retry rule for mutation requests

PR #302, opened 2026-08-06
lud03: add zero balance recommendation

### LINKED ISSUE #232: LUD-03: Provide recommendation for zero balance withdrawal response
LUD-03 should provide an explicit recommendation for what to do if a requester has no amount they could withdraw. Right now someone could choose to implement this as:

1. `minWithdrawable` and `maxWithdrawable` as `0`
2. Return an `{"status": "ERROR"}` error

While either approach should probably be handled from clients, it would be nice to have an expectation around this and reduce ambiguity.

As an aside, it might also be wise to specify that the `*Withdrawable` amounts should never be negative, unless there's a case for that that I'm not thinking of.

dni:
closes #232

### DIFF
--- a/03.md
+++ b/03.md
@@ -46,7 +46,7 @@ sequenceDiagram
     		Note left of LN Service: {<br/>"status": "ERROR", <br /> "reason": string <br/> }
     		Wallet->>Alice: notifies of unsuccessful withdrawal
     		Note over Alice: END
-	end 
+	end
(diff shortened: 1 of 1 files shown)


hsjoberg (approved):

andrerfneves:
🫡

andrerfneves (approved):

PR #303, opened 2026-08-06
LUD-XX: Add multi-rail payment options to LNURL-pay (LUD-06)

andrerfneves:
Bitcoin and Lightning payments have expanded—like it or not—across multiple rails, chains, sidechains, and offchain systems. Unless these flows are abstracted and unified from the user’s perspective, the resulting experience will be fragmented and confusing. We risk repeating the same usability problems that have constrained Bitcoin payment adoption for more than a decade.

LNURL is already deployed across virtually all Lightning-capable wallets, while most new Bitcoin payment wallets also provide Lightning interoperability. LNURL-pay is therefore a practical compatibility layer through which these payment methods can be presented as one coherent flow rather than a collection of unrelated protocols.

Each payment rail has distinct properties, trade-offs, and benefits. This proposal preserves that choice while keeping the payment experience familiar: users can continue scanning existing LUD-06 `lnurl1` QR codes or using LUD-16 Lightning Addresses, then select from the payment options supported by both the service and their wallet.

This PR introduces `paymentOptions` as a backward-compatible extension to LUD-06. Services can advertise supported payment methods, and compatible wallets can select one through the existing LNURL-pay callback flow. Users do not need to understand the technical distinctions between BOLT11, BOLT12, onchain Bitcoin, Liquid, Arkade, Bark, Spark, and future payment methods simply to complete a payment.

Wallets that do not support this extension continue using the standard BOLT11 flow unchanged. This provides an incremental path for introducing multi-rail payments without breaking the LNURL infrastructure already deployed throughout the Lightning ecosystem.

The proposal also extends LUD-21 verification semantics for payment options that do not return a BOLT11 invoice. Its scope is deliberately limited to payment-method selection; assets, currency conversion, quotes, and rail-specific fees remain outside this LUD. The initial registry covers Lightning, onchain Bitcoin, Liquid, Arkade, Bark, BOLT12, and Spark while remaining extensible to future payment methods.


shocknet-justin:
concept nack

The beauty of LN (and bolt11 specifically) is that it's the lingua franca, fake layer2's already shim in bolt11 support because they are not lightning and have no choice to expose their closed networks to Lightning's open network.  Support for these closed networks is dillution of the "LN" in LNURL and muddles the case for Lightning as the lingua franca.

Even bolt12 is positioned to obviate LNURL, doing so poorly through a Tor-like mechanism. Delivering bolt12 via LNURL is adding yet another round-trip on top of the one Bolt12 already creates, adding yet another failure point in an already failure prone protocol.

Kukks:
Implemented this in a live LNURL service ([ArkLabsHQ/lnurl-server](https://github.com/ArkLabsHQ/lnurl-server), running on mutinynet) — `paymentOptions` advertising, `paymentOption` selection, per-option `minSendable`/`maxSendable`, `paymentDestination` + `paymentURI`, and the LUD-21 verify extension. It fits well, and the `id`/`type` split earned its keep immediately: our `arkade` option is served by two different internal rails and the wallet never needs to know.

Two things came up that the spec does not cover, one of which I think is a real gap.

**A destination can expire, and there is no way to say so.**

BOLT11 carries its own expiry, so for `type: "lightning"` the question never arises. Every non-`pr` option in the registry is an address or a destination, and the spec reads as though those are payable indefinitely. For plenty of services they are not:

- A service watching an address only attributes payments for some window — ours is seven days — after which `verify` answers "not found" and the payment leaves no trace, even though the money arrived.
- A destination produced by a swap has a real deadline: the counterparty reclaims the funds if nothing happens by then.
- Anything holding a per-payment contract has its own timelocks.

Without a field for this the payer has no way to learn a deadline exists, and the failure is silent — they pay late, the money moves, and the service has no record of it. We ship an `expiresAt` (unix seconds) on the callback response as an unknown-field extension, which §"`paymentOptions` fields" explicitly permits, but an extension that every address-rail service will independently need is a sign it belongs in the spec.

Worth noting the semantics are not "the destination stops working" — in our case a late payment still reaches the user. What expires is the service's willingness to attribute it, which is exactly the thing `verify` speaks to. So it may sit more naturally as a LUD-21 concern than a rail-selection one; happy either way, but it currently has no home.

**A small note on `available`.**

The field is useful and we now emit it, but the spec does not say what distinguishes `available: false` from omitting the option. We landed on: omit when the address does not offer the rail at all (never configured, or turned off by the user), and `available: false` when it is offered but the service cannot serve it this minute. Omitting in the second case tells a payer "this address does not do lightning", and they go elsewhere instead of retrying — a materially different instruction. A sentence pinning that down would stop implementations diverging on it.

Neither blocks adoption. Thanks for pushing this forward — the split between rail selection and amount/asset denomination is the right cut, and keeping quotes out of scope made it much easier to implement.

PR #305, opened 2026-09-03
LUD-25: allow a persistent SERVICE signing key

TheCryptoDonkey:
## Why

Offline verification is now mandatory for a SERVICE, which I think is right.  Requiring `mintPubkey` to be the funding node identity makes that impossible for a backend which can fund the mint but exposes no compatible `signmessage` API, such as Spark.  Making signatures optional again would remove the property we actually need.

## What

- keep note signatures mandatory
- require `mintPubkey` to be stable, persistent and controlled by SERVICE
- prefer the funding node identity where compatible signing exists
- allow a dedicated SERVICE key where it does not
- require wallets to pin the key by full SERVICE origin and never replace a pin silently

## Trust boundary

A dedicated key cannot be corroborated from the BOLT-11 invoice.  Offline verification then proves issuance by the pinned SERVICE key, not by the funding node.  The text says that directly.

Rotation is a SHOULD NOT, with no key history.  SERVICE authenticates a note by `sha256(k1)` and never by its signature, so a rotation puts no funds at risk, only the offline attestation, and the spec's existing repair path applies: rotating such a note once re-signs it under the current key.  A published history would also have got the urgent case backwards, since the rotation that needs to happen fast is a leaked key.

## Implementation impact

LND and CLN-backed services can keep using the node identity.  A Spark-backed reference mint can load a persistent local key and produce the same signature wire format.

This PR targets the existing `lnurlcash` branch behind #301.  It does not merge or change the `luds` base branch directly.  Documentation-only change; `git diff --check` passes.

### DIFF
--- a/25.md
+++ b/25.md
@@ -176,14 +176,13 @@ A bearer note as defined above is an opaque secret: an offline recipient has no 
  {
    "tag": "withdrawRequest",
    ...
-+  "mintPubkey": string, // 33-byte compressed secp256k1 pubkey, hex
-+  "previousPubkeys": string[] // earlier mintPubkeys; present iff the key has rotated
++  "mintPubkey": string // 33-byte compressed secp256k1 pubkey, hex
  }
 ```
 
-`SERVICE` MUST rotate `mintPubkey` only deliberately. After a rotation it MUST publish every earlier `mintPubkey` under which a note may still circulate in `previousPubkeys`, for at least as long as it remembers notes and burns signed under that key. Old private keys need not be retained: verifying an existing signature needs only the public key, and new notes are signed only by the current key.
+`SERVICE` SHOULD NOT rotate `mintPubkey`. A rotation puts no note at risk, `SERVICE` authenticates a note by `sha256(k1)` and never by its signature, but it does end offline verification for every note signed under the old key. A holder restores verifiability by rotating such a note once, which re-signs it under the current key.
 
-`WALLET` MUST pin the accepted `mintPubkey` and `previousPubkeys` to the full `SERVICE` origin. A newly advertised `mintPubkey` MUST NOT silently replace a pinned one. `WALLET` MUST require explicit holder approval unless the change is authenticated by a separate continuity mechanism. Merely listing the old public key in an unsigned response preserves verification of old notes but does not prove that the old key authorised the new one.
+`WALLET` MUST pin the accepted `mintPubkey` to the full `SERVICE` origin. A newly advertised `mintPubkey` MUST NOT silently replace a pinned one, `WALLET` MUST require explicit holder approval.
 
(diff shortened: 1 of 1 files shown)


25.md
```diff
@@ -170,16 +170,21 @@ A `WALLET` that does not implement `LNURLcash` treats a received note as any oth
...
+`SERVICE` publishes its signing key as `mintPubkey` (33-byte compressed secp256k1 public key, hex) in its `withdrawRequest` JSON response. `mintPubkey` MUST be a stable key controlled by `SERVICE`. It SHOULD be the funding node identity when that backend exposes a compatible `signmessage` API, because a `WALLET` paying a BOLT-11 invoice can recover the node id from the invoice's own signature and corroborate the same identity without another trust root. A `SERVICE` whose backend cannot produce compatible signatures MAY instead use a dedicated signing key. A dedicated key MUST be persisted securely and MUST NOT be generated anew when `SERVICE` restarts. In that case offline verification proves issuance by the pinned `SERVICE` key, not by the Lightning node that funded the mint.
 
 ```diff
  {
    "tag": "withdrawRequest",
    ...
-+  "mintPubkey": string // 33-byte compressed secp256k1 pubkey, hex
++  "mintPubkey": string, // 33-byte compressed secp256k1 pubkey, hex
++  "previousPubkeys": string[] // earlier mintPubkeys; present iff the key has rotated
```
dni: i do not think there is a need for previous pubkeys.

TheCryptoDonkey: I think we need some form of rotation history if rotation is allowed.  A wallet installed after a rotation, or one restoring or receiving an older still-outstanding note, otherwise sees only the new `mintPubkey` and can never verify the old signature.  Wallet-local history helps only wallets that witnessed the change.

We could remove `previousPubkeys` if the rule instead says `mintPubkey` MUST never rotate while any note signed by it may still circulate, which is effectively forever here.  I would keep the public-key history as the minimum compatibility mechanism.  It is not self-authenticating, so the following paragraph deliberately requires explicit holder approval or a separate authenticated continuity mechanism.

dni (approved):

25.md
TheCryptoDonkey: Agreed, dropped in 4497174.

You are right that the spec already answers this: a note that lacks a usable certificate is repaired by rotating it once, which is exactly what the section below already says for freshly minted and third-party notes.  `SERVICE` authenticates by `sha256(k1)` and never by the signature, so a rotation puts no funds at risk, only the offline attestation, and one round trip restores it.  The list also got the urgent case backwards: the rotation that actually needs to happen fast is a leaked key, and a MUST-publish history with no revocation flag keeps it trusted.

What is left is a SHOULD NOT rotate, plus the pinning rule, which is the part doing the work.  lnurl-wallet#91 is updated to match: the previous-key handling and its staging UI are gone.

PR #307, opened 2026-09-07
LUD-25: retain and distinguish spent note hashes

TheCryptoDonkey:
A wallet checking a previously redeemed note by hash needs to distinguish a retained spent record from an identifier the mint never registered. The current draft explicitly groups both under unknown, so this proposal changes that contract.

Require SERVICE to retain registered hashes and their spent state. Informational lookups return `Note already spent.` for known spent notes, `Unknown note.` for unknown identifiers and `pending` for unresolved melts. Hash lookup remains optional and read-only. A confirmed unpaid melt restores outstanding value.

Hash lookup protects the spending secret; it exposes existence and spent state to someone who knows the hash. WALLET must preserve held/replacement secrets after an inconclusive hash response, including replies from older mints, and must not automatically retry unknown by revealing k1. For its own melt it should use matching settlement verification when available.

Replacement for closed draft #306. Draft for DNI review. Coordinated implementations and checks:
- Reference mint: https://github.com/lnurlcash/lnurl-mint/pull/38
- Companion vault-retirement fix (existing device protocol; no firmware change): https://github.com/lnurlcash/lnurl-wallet/pull/109
- Moneyer: https://github.com/forgesworn/moneyer/pull/43
- Conformance grader: https://github.com/lnurlcash/lnurlcash-conformance/pull/21

Validation: reviewed the informational/callback parameter boundary and pending/settlement wording; `git diff --check` passed. These are proposed protocol semantics, not a claim about deployed behaviour.


25.md
```diff
@@ -88,7 +88,11 @@ These semantics apply **only to the `callback` URL** obtained from the `withdraw
...
+
+A `WALLET` can therefore reconcile a spent note by hash without disclosing its bearer secret. Hash lookup protects the spending secret; it does not hide the note's existence or spent state from someone who already knows its hash. An unknown response alone is not proof of spending or of a completed mutation: the identifier may never have been registered, or an older `SERVICE` may not support this lookup form or distinguish spent hashes. `WALLET` MUST NOT discard held or replacement note secrets on the strength of an unknown response alone, and MUST NOT automatically retry that response by disclosing `k1`. For a melt it initiated, `WALLET` SHOULD use matching payment settlement verification where available; a pending response or transport failure is not confirmation that the melt settled.
```
dni:
```suggestion
```

dni (approved):

PR #309, opened 2026-09-15
Clarify LNURLcash address-management security boundary

TheCryptoDonkey:
## Summary

- clarify that a published `cx1` is watch-only but publicly enumerable and linkable
- separate LUD-25 receiving semantics from username allocation and lifecycle management
- state the authentication, replacement and replay properties any separate management protocol must preserve

The broader management design is now tracked in #310 rather than added as an orphan non-normative file.

This targets the existing `lnurlcash` feature branch behind PR #301, not the repository default branch.

## Validation

- `git diff --check origin/lnurlcash...HEAD`
- manually checked terminology and relative Markdown links against current `25.md`


dni:
i think this is solved by including the domain inside the adderssproofs

PR #311, opened 2026-09-16
fix: pin Part 2's address branch under its own m/139'/1' sub-purpose

TheCryptoDonkey:
The Seed & derivation formula as written roots Part 2's branch at the exact node `derive_cash_domain_node` already needs for a deterministic Part 1 secret scheme - not defined in this document, but a natural extension of the same seed, and already shipped in the reference wallet. Following the text literally collides the two: a wallet implementing both would find none of its own Part 1 notes once Part 2 derives over the same host.

This documents what the reference wallet (lnurl-wallet) already does - `m/139'/1'/d1/d2/d3/d4`, hashing key at `m/139'/1'/0` - and what lnurl-wallet#166 cross-validates against an independent client, so implementations converge on one path instead of drifting further apart.

No change to Part 1, Wallet-side ownership proofs, or anything downstream of the branch root - just the two derivation lines.


TheCryptoDonkey:
Heads up: dni's dropping the whole Part 1 seed-derivation extension from lnurl-wallet (lnurlcash/lnurl-wallet#166) rather than adding a purpose hop for Part 2 to dodge it. If that lands, there's no collision left for this PR to document - the original wording (no `1'` hop) was correct. Holding this open until #166 merges, then closing or reworking depending on what actually ships.

TheCryptoDonkey:
Confirmed stale - closing. This documented the m/139'/1' hop as necessary, but lnurl-wallet#166 drops the whole Part 1 seed-derivation extension it was protecting against, so the literal spec text (no hop) was right all along. Matching fixes already up: lnurlcash-core#15 and lnurlcash-conformance#28, both stacked on the Schnorr ck1 branches.

